PR #66, opened 2022-01-31
Update README.md

StarMKWii:
Add the 'json5' library to the dependencies list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ You need:
 
 - Python 3
 - pyjson5
+- json5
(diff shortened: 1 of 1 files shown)


StarMKWii:
> stebler — Today at 10:13
> @Star pyjson5 is already in the dependency list the potentially confusing part is that it's not called like that by package managers
> on pip pyjson5 is actually a different project
> and the one we use is just json5
> on Arch Linux it's also python-json5

Issue #89, opened 2022-01-31
No longer correct text.


Zachruff:
![image](https://user-images.githubusercontent.com/70397032/151875227-1b3f9f3f-ac3c-41bd-bc42-41c8030198ad.png)
this should be changed as you can no longer check records.

PR #91, opened 2022-02-01
Change BMG 0x7eb for it to match the menu

Lami-Lemmy:
All languages have the string in English
Solves #89


stblr:
Can you mention that you can change settings? Also you can keep everything in English because we are going to work with the Translation Project anyway.

Lami-Lemmy:
Changes pushed!

PR #158, opened 2022-02-13
Optimization: Uncap emulation speed during loading sequences

### LINKED ISSUE #157: Unlock emulation speed during load sequences

riidefi:
Drastically reduces (more than halves) loading times on Dolphin, assuming non-toaster hardware.

Closes #157

### DIFF
--- a/symbols.txt
+++ b/symbols.txt
@@ -75,6 +75,10 @@
 0x80242c18 EGG_TaskThread_request
 0x80243d6c EGG_Video_configure
 
+0x80215168 EGG_ColorFader_CT
+0x80215280 EGG_ColorFader_fadeIn
+0x802152A8 EGG_ColorFader_fadeOut
+
 0x80386000 s_systemManager
 0x80386638 MaxEntryNum
 0x8038663c FstStringStart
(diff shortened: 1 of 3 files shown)


symbols.txt
```diff
@@ -75,6 +75,9 @@
 0x80242c18 EGG_TaskThread_request
 0x80243d6c EGG_Video_configure
 
+0x802152A4 EGG_postFadeIn
+0x802152D0 EGG_postFadeOut
+
```
stblr, line 80: For these maybe keep the original symbol name and an offset, that might be preferred if we end up using them e.g. in the exception handler.

riidefi: As a comment? Or should we add a PATCH_BRANCH variant that supports offsets?

stblr: `PATCH_B(EGG_ColorFader_fadeIn + 0x24, post_fadeIn);` already works.

riidefi: Gotcha. Didn't realize function pointer arithmetic like that was valid here.

riidefi:
Resolved

stblr (approved):

PR #181, opened 2022-02-19
Complete the section layouts of the '.dol' and '.rel' modules for all regions

StarMKWii:


port.py
```diff
@@ -497,10 +586,14 @@ def port(region, address):
...
-    write_symbol(out_file, f'{name}_end', dst_binary.end)
-    out_file.write('\n')
+for section in SRC_BINARIES[args.region]['dol'].sections:
+    write_symbol(out_file, f'dol_{section.name}_start', section.start)
+    write_symbol(out_file, f'dol_{section.name}_end', section.end)
+out_file.write('\n')
+
+write_symbol(out_file, 'rel_start', DST_BINARIES[args.region]['rel'].start)
+write_symbol(out_file, 'rel_end', DST_BINARIES[args.region]['rel'].end)
+out_file.write('\n')
```
stblr, line 596: We will need symbols for all sections of both the dol and the rel, as well as binary start and end symbols (for the dol we can keep them even if they are the same as the first/last section, to make the intent clearer). For the rel, they will need to be ported to the mkw-sp address mapping.

stblr (changes requested):

port.py
StarMKWii: These requirements should now be satisfied.

stblr:
The code looks ok but there are some conflicts with the main branch.

stblr:
Applied in #187.

PR #198, opened 2022-02-25
Add NetStorageClient.c

riidefi:


payload/sp/Tcp.h
```diff
@@ -0,0 +1,23 @@
...
+
+void TcpSocket_Init(TcpSocket *sock);
+
+bool TcpSocket_isConnected(const TcpSocket *sock);
+bool TcpSocket_tryConnectIpv4(
+        TcpSocket *sock, u8 ip_a, u8 ip_b, u8 ip_c, u8 ip_d, u32 port);
+void TcpSocket_disconnect(TcpSocket *sock);
+
+bool TcpSocket_receiveBytes_(TcpSocket *sock, void *buf, u32 len);
+bool TcpSocket_sendBytes_(TcpSocket *sock, const void *buf, u32 len);
```
stblr, line 23: Is there a reason for different naming (uppercase, trailing underscore)?

riidefi: Fixed

PR #203, opened 2022-03-01
Add network-attached storage support

riidefi:


payload/sp/WideUtil.h
```diff
@@ -0,0 +1,21 @@
...
+    }
+}
+static inline void Util_toUtf8(char *dst, u32 dst_max, const wchar_t *src, u32 src_max) {
+    const u32 min_bound = MIN(dst_max, src_max);
+    const u32 len = wcslen(src /*, min_bound*/);
+
+    for (u32 i = 0; i < MIN(min_bound, len); ++i) {
+        dst[i] = (char)src[i];
+    }
+}
```
stblr, line 21: This is ASCII-only, is that intended?

payload/game/system/RootScene.c
```diff
@@ -1,103 +1,8 @@
...
-}
 
 // Memory layout isn't randomized here
-void RootScene_allocateEx(RootScene *scn) {
-    RootScene_initNet(scn);
+void RootScene_allocateEx(RootScene * UNUSED(scn)) {
+    
 }
 
 PATCH_B(RootScene_allocate + 0xb64, RootScene_allocateEx);
```
stblr, line 8: This can be removed I think.

stblr (changes requested):

payload/sp/WideUtil.h
riidefi: Does the game ever read any non-ASCII paths? We could change the API to just forward UTF16 paths to the server.

riidefi:
Resolved

payload/sp/WideUtil.h
stblr: No, that would only be useful for ghost files which aren't supported with NetStorage.

stblr (approved):

PR #216, opened 2022-03-07
NetStorageClient: Support compiling on PC

riidefi:
This will allow for fuzzing/better debugging.

### DIFF
--- a/payload/sp/Bytes.h
+++ b/payload/sp/Bytes.h
@@ -2,6 +2,70 @@
 
 #include <Common.h>
 
+static inline u32 swap32(u32 v);
+static inline u16 swap16(u16 v);
+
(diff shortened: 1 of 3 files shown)


payload/sp/Bytes.h
```diff
@@ -2,6 +2,70 @@
...
+#else
+static inline u32 swap32(u32 v) {
+    return _BSWAP_32(v);
+}
+static inline u16 swap16(u16 v) {
+    return _BSWAP_16(v);
+}
+#endif
+
+#ifdef PLATFORM_EMULATOR
```
stblr, line 37: This should be a little endian check

stblr (changes requested):

riidefi:
Resolved both

stblr (approved):

PR #229, opened 2022-03-21
Add check for MKW disk

vabold:


stblr (approved):
We should probably try to eject the disk in that case.

PR #232, opened 2022-03-24
IOS: Log opens

riidefi:


payload/revolution/ios.c
```diff
@@ -0,0 +1,36 @@
...
+// ios.S
+extern s32 real_IOS_Open(const char *path, u32 flags);
+extern s32 real_IOS_OpenAsync(const char *path, u32 flags, void *cb, void *userdata);
+
+static bool CanIOSOpen(const char *path, u32 flags) {
+    (void)path;
+    (void)flags;
+
+    return true;
+}
```
stblr, line 13: Are there devices we will want to block?

stblr (approved):

PR #301, opened 2022-05-27
Add the 'const' qualifier to the 'size' parameter of the function 'Stack_AllocFromMEM1ArenaHi'

StarMKWii:


stblr (approved):
I'm going to merge it but I'm not really convinced that using `const` on non-pointer types is worth the verbosity.

riidefi:
> I'm going to merge it but I'm not really convinced that using `const` on non-pointer types is worth the verbosity.

The C++ Core Guidelines seem to hold the same position:
> **Exception** Function parameters passed by value are rarely mutated, but also rarely declared const. To avoid confusion and lots of false positives, don’t enforce this rule for function parameters.
> ```cpp
> void f(const char* const p); // pedantic
> void g(const int i) { ... }  // pedantic
> ```
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconst-immutable

PR #302, opened 2022-05-31
Move the definition of the variable 'mem1_arena_hi' into the linker script file

StarMKWii:


stblr (commented):
What about making it a `u8 *` and removing the temporary?

StarMKWii:
> What about making it a `u8 *` and removing the temporary?

Addressed in commit [4ac4097](https://github.com/stblr/mkw-sp/pull/302/commits/4ac409714f8ea04e58db51b67c3c377d363b6b32).

stblr (approved):

PR #318, opened 2022-06-27
Parse update response messages in Protobuf format

StarMKWii:


stblr (approved):
Merging this now, will test later.

PR #390, opened 2022-08-19
Store log files in the directory '/mkw-sp/logs'

StarMKWii:


stblr:
Is the file name change going to be its own PR?

payload/sp/storage/LogFile.cc
```diff
@@ -19,7 +19,8 @@ static OSThread thread;
 static u8 stack[4096];
 
 static void *Run(void *UNUSED(arg)) {
-    auto file = Storage::Open(L"/mkw-sp/log.txt", "w");
+    Storage::CreateDir(L"/mkw-sp/logs", false);
```
stblr, line 22: Nitpick: it doesn't matter since we aren't checking the return value, but the second arg should be `true`.

StarMKWii:
> Is the file name change going to be its own PR?

That is correct.

payload/sp/storage/LogFile.cc
StarMKWii: Addressed in commit [babbee4](https://github.com/stblr/mkw-sp/pull/390/commits/babbee4e4c0d0e66071181af36c3d7ecb6f24af2).

stblr (approved):

PR #402, opened 2022-08-31
Mirror tts

henrymkw:
Mirror mode TTs implementation currently does not work.


henrymkw:
Current issue is that going from mirror enabled to mirror disable causes what happens in the screenshot.
https://gyazo.com/c3209cb83e8d660a20ed79b2ba854f75

vabold:
> Current issue is that going from mirror enabled to mirror disable causes what happens in the screenshot. https://gyazo.com/c3209cb83e8d660a20ed79b2ba854f75

This requires working with nw4r (specifically a BRRES wrapper), which requires selling your soul to the devil. Or riidefi. Either one will work.

Rii claims there's an easy solution here involving unflipping the material in the dtor at 0x8055B7F8 PAL, but I'm too tired to do anything related to nw4r.

PR #410, opened 2022-09-19
Use the optimized version of the function 'memcpy'

MikeIsAStar:
Aims to resolve issue https://github.com/stblr/mkw-sp/issues/149.

### DIFF
--- a/payload/platform/string.c
+++ b/payload/platform/string.c
@@ -1,7 +1,7 @@
 #include <Common.h>
 #include <string.h>
 
-extern void *memcpy_slow(void *dest, const void *src, size_t n);
+void *memcpy_slow(void *dest, const void *src, size_t n);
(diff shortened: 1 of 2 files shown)


payload/platform/string.c
```diff
@@ -1,6 +1,11 @@
 #include <Common.h>
 #include <string.h>
 
+extern void *memcpy_slow(void *dest, const void *src, size_t n);
```
stblr, line 4: No need for `extern` here.

MikeIsAStar: Resolved in commit [6d21df6](https://github.com/stblr/mkw-sp/pull/410/commits/6d21df659831f9c4dfe04d017129987c3ecf809c).

stblr (approved):

PR #439, opened 2022-10-26
Add CourseMap, additional gx headers

riidefi:


stblr (approved):
KMP structures right when I need them!

PR #452, opened 2022-11-02
Allow selecting the same character in splitscreen multiplayer

TheLordScruffy:
Some memory optimizations will need to be made to get the 4 player menu to load, as right now it's running out of memory


payload/game/ui/model/CharacterModelManager.S
```diff
@@ -0,0 +1,234 @@
...
+    lwz     r3, 0x4C(r27)
+    cmpwi   r3, 0
+    bne-    alreadyAllocated
+
+    // Allocate new pigtails
+    lbz     r3, 8(r27)
+    mulli   r3, r3, 0x58
+    bl      new
+    stw     r3, 0x4C(r27)
+    b       alreadyAllocated
```
vabold, line 93: Possibly rename `alreadyAllocated` to `allocated` and remove unnecessary instruction.

stblr (commented):
LGTM if @vabold's suggestions are addressed.

payload/game/ui/ctrl/CtrlMenuCharacterSelect.cc
```diff
@@ -7,36 +7,19 @@ void CtrlMenuCharacterSelect::ButtonDriver::initSelf() {
...
-        setMessage("ok_text_2p", 2555 + 1);
-    }
-
-    if (auto *pane = m_mainLayout.findPaneByName("ok_text_3p")) {
-        if (auto *material = pane->getMaterial()) {
-            material->tevColors[1] = { 255, 255, 255, 255 };
-        }
-        setMessage("ok_text_3p", 2555 + 2);
-    }
+    const char *panes[4] = {"ok_text_1p", "ok_text_2p", "ok_text_3p", "ok_text_4p"};
```
vabold, line 14: Replace with snprintf in the loop.

TheLordScruffy: I don't agree with this

vabold: I can agree to disagree. `panes` should be static to avoid conflict with other UI files.

riidefi: I think this PR's way is fine, though the following shorthand should also work:
```cpp
for (const char *pane : {"ok_text_1p", "ok_text_2p", "ok_text_3p", "ok_text_4p"}) {
    ...
}
```
provided `initializer_list` is included.

TheLordScruffy: that method doesn't give the index, which is needed for the message ID

riidefi: Or I suppose `auto [i, name] : std::views::enumerate({"ok_text_1p", "ok_text_2p", "ok_text_3p", "ok_text_4p"})` to keep `i` but that's far more cumbersome...

riidefi: Yea this approach is fine. `std::size(panes)` could replace `4` if you want.

riidefi: > `panes` might need to be static to avoid conflict with other UI files - maybe stebler or riidefi can tell me if I'm wrong.
@vabold Shouldn't be the case since this is inside a function. We only need to worry about names conflicting on global scope.

vabold:
LGTM

stblr (approved):

PR #456, opened 2022-11-04
Halt code execution if heap allocation fails

MikeIsAStar:


vabold:
I think we need something a bit more unique for the message, since we already use `Out of memory!` for a different panic. We could possibly use `OOM - heap allocation failed!` and `OOM - memory allocation failed!`.

MikeIsAStar:
> I think we need something a bit more unique for the message, since we already use `Out of memory!` for a different panic. We could possibly use `OOM - heap allocation failed!` and `OOM - memory allocation failed!`.

Resolved in commit [09d94ad](https://github.com/stblr/mkw-sp/pull/456/commits/09d94adca1d7c28553cbffc5ec91500d876c4651).

MikeIsAStar:
This was deemed unnecessary by stebler.

PR #462, opened 2022-11-06
Keyboard support for DirectConnectionPage

riidefi:
Use number, backspace and enter keys to interact with the UI.


payload/game/ui/DirectConnectionPage.cc
```diff
@@ -145,4 +164,44 @@ void DirectConnectionPage::onBadConnectCode([[maybe_unused]] MessagePage *messag
...
 }
 
+void DirectConnectionPage::onKeyCode(char key, u8 mods) {
+    if (key == kSimpleEvent_Backspace) {
+        if (m_okButton.isSelected()) {
+            m_digitButtons[1].select(0);
+        }
+        if (mods & kSimpleMods_CTRL) {
+            // TODO: this should only clear everything to the left of the cursor. Right now this
+            // clears everything.
```
stblr, line 174: In my opinion this is reasonable because this matches what the reset button does.

payload/game/ui/NumericEditBox.hh
```diff
@@ -18,6 +18,10 @@ public:
     bool isEmpty();
     bool isFull();
     u64 getNumber();
+    
+    // Presumably localPlayerId
```
stblr, line 22: Indeed. There is one more argument in r5 (`s32`) but I don't know what it does, as it seems to be unused most of the time.

payload/game/ui/DirectConnectionPage.cc
riidefi: I suppose in most text editors CTRL+BACKSPACE would clear to the end of the word; in this case, until the hyphen? Regardless it is probably fine as-is. Updating the comment.

payload/game/ui/NumericEditBox.hh
riidefi: Resolved

payload/game/ui/DirectConnectionPage.cc
```diff
@@ -145,4 +166,43 @@ void DirectConnectionPage::onBadConnectCode([[maybe_unused]] MessagePage *messag
     reinterpret_cast<MenuMessagePage *>(messagePage)->m_replacement = PageId::OnlineTop;
 }
 
+void DirectConnectionPage::onKeyCode(char key, u8 mods) {
```
stblr, line 169: For this, I think that we could even call `PushButton::onFront` 805be358 to have the sound and animation.

riidefi: Yep, that works. Changed

stblr (approved):

PR #491, opened 2022-11-21
Add a default value for the 'Region Line Color' setting

MikeIsAStar:


assets/message/MenuSP_U.bmg.json5
```diff
@@ -1163,6 +1163,14 @@
         "font": "regular",
         "string": "Speedometer displays the speed the vehicle moves vertically.",
     },
+    "10332": {
+        "font": "regular",
+        "string": "Default",
+    },
+    "10333": {
+        "font": "regular",
+        "string": "The color of the line under your name will be\ndetermined by the game.",
```
Zachruff, line 1172: better wording might be: 

The color of the line under your name will be\nset automatically based on your region settings.

MikeIsAStar: > The color of the line under your name will be\nset automatically based on your region settings.

That statement is not entirely accurate.

PR #498, opened 2022-12-05
Tell the user how to set MKW as the default ISO

### LINKED ISSUE #468: Explain how to insert the disc in the loader if Dolphin is detected
We often get this question in support channels.

Lami-Lemmy:
Will only show up if dolphin is detected.
Resolves #468

### DIFF
--- a/common/IOS.cc
+++ b/common/IOS.cc
@@ -140,7 +140,7 @@ u32 armCode[] = {
 #endif
 // clang-format on
 
-static bool IsDolphin() {
+bool IsDolphin() {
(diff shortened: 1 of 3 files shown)


vabold:
Consider removing the initial disc change prompt on Dolphin, or at least modifying its wording to reflect more about what's required on the user's part.

vabold (approved):

Naim2000:
But setting as default ISO doesn't insert the disc. You have to restart the emulation or select change disc as well. :thinking:

PR #506, opened 2022-12-12
Load clients into the race with custom gamemode/player type

vabold:


vabold:
Can someone review the ASM and make sure the patch macros are correct?

payload/game/kart/KartObject.S
```diff
@@ -1,5 +1,22 @@
 #include <Common.S>
 
+PATCH_BL_START(KartObject_Create, 0x34)
```
Gaberboo, line 3: Since you don't use LR, PATCH_B makes more sense. All other patch macros are good.

vabold: Resolved in https://github.com/stblr/mkw-sp/commit/60a4da2d8f12f2944134fa55299f2491470a7850.

vabold:
@stblr implemented it separately into the "rollback" branch.

PR #524, opened 2022-12-28
Updated Direct Connection Code Tool

TheNinjaKingOW:
If no arguments are provided, the script will prompt the user to input the arguments.

If only 1 or 2 arguments are missing, it will provide an error.

### DIFF
--- a/tools/directconnectioncode/directconnectioncode.py
+++ b/tools/directconnectioncode/directconnectioncode.py
@@ -33,8 +33,7 @@ def generate(ip: str, port: int, passcode: int):
     direct_code = direct_code[:6] + '-' + \
         direct_code[6:12] + '-' + direct_code[-6:]
 
-    print("Your direct connection code is \"" + direct_code + "\".")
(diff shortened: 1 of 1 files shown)


tools/directconnectioncode/directconnectioncode.py
```diff
@@ -33,14 +33,24 @@ def generate(ip: str, port: int, passcode: int):
     direct_code = direct_code[:6] + '-' + \
         direct_code[6:12] + '-' + direct_code[-6:]
 
-    print(direct_code)
+    return direct_code
```
vabold, line 36: Because it now returns a string, the return typehint should be `str` to reflect this.

tools/directconnectioncode/directconnectioncode.py
```diff
@@ -1,4 +1,4 @@
-from argparse import ArgumentParser
+import argparse
```
vabold, line 1: Unnecessary import change?

vabold (changes requested):

PR #532, opened 2023-01-09
Add EGG::Disposer

riidefi:


vabold:
Any reason for `m_containHeap != nullptr` instead of `m_containHeap`?

riidefi:
> Any reason for `m_containHeap != nullptr` instead of `m_containHeap`?

https://abseil.io/tips/141

stblr:
Do we still want this now that a simpler approach was chosen in #536?

riidefi:
It can't hurt to have this header available moving forward.

stblr (approved):

PR #534, opened 2023-01-09
Allow disabling of the speedometer

### LINKED ISSUE #527: There exists no option to disable the speedometer

GnomedDev:
Adds an option to disable the speedometer's rendering.

Closes #527.

### DIFF
--- a/assets/message/MenuSP_U.bmg.json5
+++ b/assets/message/MenuSP_U.bmg.json5
@@ -1217,6 +1217,12 @@
     "20043": {
         string: "Take me back to safety!",
     },
+    "20044": {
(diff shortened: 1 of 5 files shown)


stblr:
For this it should be possible to toggle the setting without a restart.

stblr:
The way to do it would be to use an animation for CtrlRaceLap.

GnomedDev:
I haven't a clue how the animation system works, so that would have to be implemented in a follow up PR. As is, this PR is entirely usable and implements the request, someone else could implement an animation based approach in the future if they want.

stblr (approved):

PR #540, opened 2023-01-15
Get battle mode playable again

GnomedDev:
Currently, battle mode is in quite a sorry state with

- [x] Kart Select not rendering any kart other than standard kart and bike
- [x] A crash on loading battle demo in `getPlayerIdx` due to a null pointer
- [x] The battle demo showing VS UI
- [ ] The battle demo not showing wins/losses
- [ ] Current balloons popped/coins collected counter stuck at `0-0`
- [x] MKW-SP crashing after battle mode ends due to results screen failing to load.

This "tracking PR" aims to fix all these issues, and will be marked ready once everything is working to at least vanilla levels.


GnomedDev:
Fixing kart select is going to be harder than I thought. It is currently broken due to the page merging effort (the UI flow has been changed to mimic VS, then branch back to battle mode when loading demo) as the normal `VehicleSelectPage` doesn't contain the code `onActivate` for calling `0x80860bf0` and setting the `LayoutUiControl`s up correctly, which even if that is fixed will probably only solve the missing thumbnails and not the missing model and spinning animation.

payload/game/enemy/EnemyManager.cc
```diff
@@ -8,6 +8,10 @@ EnemyManager *EnemyManager::Instance() {
     return s_instance;
 }
 
+bool EnemyManager::isOfflineBattle() {
+    return false;
+}
```
vabold, line 13: Static methods should be PascalCase and not camelCase.

payload/game/enemy/EnemyManager.hh
```diff
@@ -11,6 +11,7 @@ public:
     static EnemyManager *Instance();
 
 private:
+    REPLACE static bool isOfflineBattle(void);
```
vabold, line 14: Static methods should be defined below non-static members.

payload/game/system/RaceConfig.cc
```diff
@@ -28,27 +28,45 @@ u8 (&RaceConfig::ghostBuffers())[2][11][0x2800] {
...
+
+    switch (m_menuScenario.gameMode) {
+    case RaceConfig::GameMode::OfflineVS:
+        setting = saveManager->getSetting<SP::ClientSettings::Setting::VSClass>();
+        break;
+    case RaceConfig::GameMode::OfflineBT:
+        setting = saveManager->getSetting<SP::ClientSettings::Setting::BTClass>();
+        break;
+    default:
+        assert(false);
```
vabold, line 46: I would add a `SP_LOG` call directly above the assert so it's a little clearer what the problem is.

assets/message/MenuSP_U.bmg.json5
```diff
@@ -1278,6 +1278,26 @@
...
+        "string": "Races are always 100cc.",
+    },
+    "10363": {
+        "font": "regular",
+        "string": "50cc"
+    },
+    "10364": {
+        "font": "regular",
+        "string": "100cc"
+    },
```
vabold, line 1300: If this is for battle mode, it should be "Battles", not "Races".

payload/game/ui/page/ResultTeamTotalPage.cc
```diff
@@ -7,15 +7,20 @@
...
 
-PageId ResultTeamVSTotalPage::getReplacement() {
-    return PageId::AfterVsMenu;
+PageId ResultTeamTotalPage::getReplacement() {
+    auto raceScenario = System::RaceConfig::Instance()->raceScenario();
+    if (raceScenario.isBattle()) {
+        return PageId::AfterBTMenu;
+    } else {
+        return PageId::AfterVsMenu;
+    }
```
vabold, line 20: I would rewrite this to be one line, but you don't have to.
`return raceScenario.isBattle() ? PageId::AfterBtMenu : PageId::AfterVsMenu;`

payload/game/util/Registry.cc
```diff
@@ -4,6 +4,12 @@
 
 namespace Registry {
 
+// This disables trying to load red/blue thumbnails and models
+// for all the vehicles in the vehicle select screen
+u32 UseBattleRenders() {
+    return false;
+}
```
vabold, line 11: This returns a boolean.

vabold (changes requested):

assets/message/Menu_E.bmg.json5
```diff
@@ -1085,7 +1085,7 @@
     },
     "3400": {
         "font": "regular",
-        "string": "Rules",
+        "string": "Settings",
```
stblr, line 1088: How hard would it be to use new ids?

symbols.txt
```diff
@@ -1218,6 +1219,7 @@
 0x8072624c _ZN5Enemy9KartEnemy5isCpuEv
 0x8073942c _ZN5Enemy12EnemyManager4calcEv
 0x807397bc _ZNK5Enemy12EnemyManager7hasCpusEv
+0x80739740 _ZN5Enemy12EnemyManager15IsOfflineBattleEv
```
stblr, line 1222: Please keep them ordered.

assets/message/Menu_E.bmg.json5
```diff
@@ -1349,7 +1349,7 @@
     },
     "3475": {
         "font": "regular",
-        "string": "Set various battle rules.",
+        "string": "Change settings for this licence.",
```
stblr, line 1352: Same

payload/sp/settings/ClientSettings.hh
```diff
@@ -73,11 +79,47 @@ enum class Category {
...
     TA,
     VS,
+    BT,
     Room,
     License,
     DebugOverlay,
 };
 
+// Shared in VS, BT, and/or Room
+enum class Classes {
```
stblr, line 89: Why plural? Because of the "mixed" option?

payload/game/ui/page/BattleModeSelectPage.hh
```diff
@@ -0,0 +1,15 @@
...
+#include "game/ui/page/MenuPage.hh"
+
+namespace UI {
+
+class BattleModeSelectPage: public MenuPage {
+private:
+    REPLACE void onButtonFront(const PushButton *button);
+    u8 _0[0x6c4 - 0x430];
+};
+
```
stblr, line 12: No blank line here.

payload/game/ui/page/BattleModeSelectPage.cc
```diff
@@ -0,0 +1,40 @@
+#include "game/ui/page/BattleModeSelectPage.hh"
+#include "game/ui/SectionManager.hh"
+#include "game/ui/SettingsPage.hh"
+#include "game/system/RaceConfig.hh"
+
+extern "C" {
+#include <revolution.h>
+#include <sp/StackTrace.h>
+}
```
stblr: https://github.com/stblr/mkw-sp/issues/245#issuecomment-1213490665

stblr (changes requested):

payload/game/ui/Section.cc
```diff
@@ -299,6 +303,11 @@ void Section::addActivePage(PageId pageId) {
 void Section::addPages(SectionId id) {
     SP_LOG("&7DEBUG: Constructing section %u (0x%x)", id, id);
 
+    // Pages may use the Section's THPManager, however the
+    // stock game is unreliable in checking if it is initalised
+    // before using it. We initialise it here to avoid crashes.
+    loadTHP();
```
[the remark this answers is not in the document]
GnomedDev: 256 bytes of heap, if it isn't already initialized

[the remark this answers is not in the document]
stblr: Actually, isn't it more like 3 MiB (in 80529ac4)?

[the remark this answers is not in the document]
GnomedDev: oh huh, but it should be loaded for most sections anyway, what do you suggest I do?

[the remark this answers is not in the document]
stblr: The sections with the globe background (which uses tons of memory), race sections as well as other sections like license settings don't use them. I would suggest to just use `REPLACE`/`REPLACED` on the `onInit` function of offending pages.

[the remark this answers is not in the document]
GnomedDev: The problem is that as more `Page`s are rewritten or removed they may forget to load the `THPManager` as it isn't required for the Page, as happened in this example, as Nintendo didn't load it for all the `Page`s it used. I'd prefer a blacklist of memory intensive pages.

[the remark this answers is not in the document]
stblr: A blacklist would have the same problem as the current "whitelist", except that it would silently waste an obscene amount of memory or OOM at a random place.

payload/game/system/RaceConfig.cc
```diff
@@ -66,7 +76,6 @@ void RaceConfig::ConfigurePlayers(Scenario &scenario, u32 screenCount) {
 
     for (u32 i = 0; i < 12; i++) {
         if (scenario.players[i].type == Player::Type::Local) {
-            assert(screenId < screenCount);
```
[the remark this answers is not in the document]
GnomedDev: That assert fired on starting split screen matches, and removing it didn't break anything (tested all screen counts and vs single)

assets/message/Menu_E.bmg.json5
GnomedDev: Would need a rewrite of BattleModeSelectPage::onInit, or onActivate, not sure. I don't see what's wrong with just replacing the old text as this would also change if that ID was reused.

stblr: The issue is that for other languages, you get an incorrect text which is worse than an untranslated one. I _think_ that using the existing id and putting it in MenuSP_U would work but that would need to be tested.

stblr (approved):

PR #541, opened 2023-01-15
Rust CI

Lami-Lemmy:
This adds an action which build the rust tools, but only if the tools are modified. It also adds a filter to the trigger for the mod's action so that it does not build SP itself if a commit only changes the tools.

~~As of now, `sign` and `updateserver` fail to build as the `passterm` dependency fails to compile. I have not tested building them on a local machine, though, and so I do not know if the issue is with the CI or the dependency.~~ `passterm` has been updated, and both tools now compile.

### DIFF
--- a/.github/workflows/build_sp.yml
+++ b/.github/workflows/build_sp.yml
@@ -1,4 +1,4 @@
-name: Build
+name: Build MKW-SP
 
 on: [push, pull_request]
 
--- a/.github/workflows/build_sp.yml
+++ b/.github/workflows/build_sp.yml
@@ -4,9 +4,11 @@ on:
   push:
     paths-ignore:
       - tools/
+      - .github/workflows/build_tools.yml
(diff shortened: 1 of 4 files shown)


Lami-Lemmy:
Should be good now

.github/workflows/build_sp.yml
```diff
@@ -1,6 +1,12 @@
...
+name: Build MKW-SP
 
-on: [push, pull_request]
+on: 
+  push:
+    paths-ignore:
+      - tools/
+  pull_request:
+    paths-ignore:
+      - tools/
```
stblr, line 10: Shouldn't `.github/workflows/build_tools.yml` be ignored here?

Lami-Lemmy: That was an oversight, fixed.

stblr (approved):

PR #548, opened 2023-01-21
gameserver: Implement chat

GnomedDev:
Added support for forwarding the Comment message, while properly adding anyhow, tracing, and documentation for helping the development experience.

### DIFF
--- a/tools/gameserver/Cargo.toml
+++ b/tools/gameserver/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.0.0"
 edition = "2021"
 
 [dependencies]
+anyhow = "1.0.68"
 libhydrogen = "0.4"
 prost = "0.11"
 slab = "0.4.7"
--- a/tools/gameserver/Cargo.toml
+++ b/tools/gameserver/Cargo.toml
@@ -8,6 +8,8 @@ libhydrogen = "0.4"
 prost = "0.11"
 slab = "0.4.7"
 tokio = { version = "~1.20", features = ["rt-multi-thread", "io-util", "net", "macros", "sync"] }
+tracing = "0.1.37"
+tracing-subscriber = { version = "0.3.16", features = ["fmt", "env-filter"] }
 
 [build-dependencies]
 prost-build = "0.11"
--- a/tools/gameserver/src/client.rs
+++ b/tools/gameserver/src/client.rs
@@ -7,15 +7,7 @@ use tokio::sync::broadcast::error::RecvError as BroadcastRecvError;
 use crate::async_stream::AsyncStream;
 use crate::event::Event;
 use crate::request::{JoinResponse, Request};
-use crate::room_protocol::{
-    room_event,
-    room_request,
-
-    RoomRequest as RoomRequestOpt,
-    room_request::Request as RoomRequest,
-
-    RoomEvent as RoomEventOpt,
-};
+use crate::room_protocol::*;
 use crate::server::ClientKey;
 
 /// Represents a MKW-SP client connected to the server.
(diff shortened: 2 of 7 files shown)


tools/gameserver/src/client.rs
```diff
@@ -6,7 +6,15 @@ use tokio::sync::{broadcast, mpsc, oneshot};
...
-use crate::room_protocol::{room_request, RoomRequest};
+use crate::room_protocol::{
+    room_event,
+    room_request,
+
+    RoomRequest as RoomRequestOpt,
+    room_request::Request as RoomRequest,
+
+    RoomEvent as RoomEventOpt,
+};
```
stblr, line 17: The aliases are fine, but I would prefer if they were used everywhere and not just in this file.

tools/gameserver/src/client.rs
```diff
@@ -82,14 +90,29 @@ impl Client {
...
+        match request {
+            RoomRequest::Comment(room_request::Comment {message_id}) => {
+                let event = room_event::Comment {
+                    player_id: self.client_key.inner as u32,
+                    message_id,
+                };
+
+                self.tx.send(Request::Comment {inner: event}).await?;
+            },
+            _ => unimplemented!("Request type not implemented!")
```
stblr, line 113: This allows the client to crash the server. We rather want the default case to disconnect the client (e.g.  if a Join request is received).

tools/gameserver/src/server.rs
```diff
@@ -144,7 +145,7 @@ impl Server {
 
 #[derive(Debug)]
 pub struct ClientKey {
-    inner: usize,
+    pub inner: usize,
```
stblr, line 148: I would suggest to add a getter instead, as the value should never be changed after construction.

tools/gameserver/src/server.rs
```diff
@@ -119,6 +118,9 @@ impl Server {
                 let response = JoinResponse { rx, client_key, events };
                 let _ = join_tx.send(response);
             },
+            Request::Comment {inner} => {
+                self.tx.send(Event::Forward { inner: room_event::Event::Comment(inner) });
```
stblr, line 122: Please explicitly ignore the error here.

tools/gameserver/src/client.rs
```diff
@@ -1,34 +1,44 @@
...
 use crate::server::ClientKey;
 
+/// Represents a MKW-SP client connected to the server.
 #[derive(Debug)]
 pub struct Client {
+    /// The stream used to communicate with the client.
     stream: AsyncStream,
+    /// The channel to send [`Request`]s to the server.
     tx: mpsc::Sender<Request>,
+    /// The channel to recieve broadcasted [`Event`]s from the server.
```
stblr, line 27: receive

tools/gameserver/Cargo.toml
```diff
@@ -4,10 +4,13 @@ version = "0.0.0"
 edition = "2021"
 
 [dependencies]
+anyhow = "1.0.68"
```
stblr, line 7: What's the best practice here, since libhydrogen re-exports it?

stblr (changes requested):

stblr:
There is a leftover `error.rs` file.

tools/gameserver/Cargo.toml
GnomedDev: Adding the dependency to our project will just mean that as long as anyhow doesn't go above 1.x, it just allows our crate access to the full anyhow crate instead of just the macro that libhydrogen re-exports. It also means that libhydrogen can have a breaking upgrade which removes anyhow usage without breaking our crate while upgrading the dependency.

stblr (approved):

PR #566, opened 2023-02-04
Implement unverified device ID logins

GnomedDev:
This implements the required IOS call to fetch the device ID, then sends this information to the matchmaking server if it is not the default dolphin ID.

On the server side, it adds a migration to the DB which sets up the schema better, to allow for user creation to actually occur (this was untested previously).

This PR does not verify if the console is lying about it's device ID, and is a foundation for this verification to be built on by someone other than me.

### DIFF
--- a/payload/game/ui/RandomMatchingPage.cc
+++ b/payload/game/ui/RandomMatchingPage.cc
@@ -1,16 +1,12 @@
 #include "RandomMatchingPage.hh"
 
-#include <game/system/SaveManager.hh>
-#include <game/ui/SectionManager.hh>
-#include <sp/cs/RoomClient.hh>
+#include "game/system/SaveManager.hh"
+#include "game/ui/SectionManager.hh"
 
 #include <protobuf/Matchmaking.pb.h>
+#include <sp/cs/RoomClient.hh>
 #include <vendor/nanopb/pb_decode.h>
 #include <vendor/nanopb/pb_encode.h>
-
-extern "C" {
-#include <revolution/es.h>
-}
 
 namespace UI {
 
--- a/payload/game/ui/RandomMatchingPage.cc
+++ b/payload/game/ui/RandomMatchingPage.cc
@@ -1,12 +1,20 @@
 #include "RandomMatchingPage.hh"
 
+#include <game/system/SaveManager.hh>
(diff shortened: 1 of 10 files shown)


payload/game/ui/RandomMatchingPage.cc
```diff
@@ -1,13 +1,21 @@
...
+#include <game/ui/SectionManager.hh>
 #include <sp/cs/RoomClient.hh>
 
 #include <protobuf/Matchmaking.pb.h>
 #include <vendor/nanopb/pb_decode.h>
 #include <vendor/nanopb/pb_encode.h>
 
+extern "C" {
+#include <revolution/es.h>
+}
```
stblr, line 13:
```suggestion
#include "game/system/SaveManager.hh"
#include "game/ui/SectionManager.hh"

#include <protobuf/Matchmaking.pb.h>
extern "C" {
#include <revolution/es.h>
}
#include <sp/cs/RoomClient.hh>
#include <vendor/nanopb/pb_decode.h>
#include <vendor/nanopb/pb_encode.h>
```

payload/game/ui/RandomMatchingPage.cc
```diff
@@ -55,34 +69,74 @@ void RandomMatchingPage::afterCalc() {
...
     response.which_message = CTSMessage_login_tag;
-    response.message.login.has_client_id = false;
+    if (deviceId == DOLPHIN_DEFAULT_DEVICE_ID) {
+        response.message.login.has_client_id = false;
+        m_state = State::WaitForResponse;
+    } else {
+        response.message.login.has_client_id = true;
+        response.message.login.client_id = LoggedInId{
+            device : deviceId,
+            licence : *System::SaveManager::Instance()->spCurrentLicense(),
```
stblr, line 86: How will multiple devices using the same certificate be handled?

GnomedDev: How would multiple devices have the same certificate?

stblr: For instance if you import your Wii NAND in Dolphin and connect with both at the same time.

stblr (approved):

PR #567, opened 2023-02-04
Online race rework

stblr:


tools/gameserver/Cargo.toml
```diff
@@ -5,10 +5,10 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1.0.68"
+dlv-list = "0.5.0"
```
GnomedDev, line 8: What's the reason for swapping slab out for this?

tools/gameserver/src/room.rs
```diff
@@ -1,256 +1,370 @@
...
+                    }
+                    match self.handle_select_request(client_index, request, gamemode) {
+                        Ok(Some(gamemode)) => return Ok(Some(gamemode)),
+                        Err(e) => {
+                            // TODO handle
+                        },
+                        _ => (),
+                    }
+                },
+                else => return Ok(None),
```
GnomedDev, line 138:
```suggestion
                _ => return Ok(None),
```

tools/gameserver/src/room.rs
```diff
@@ -1,256 +1,370 @@
...
-                inner,
-            } => {
-                if self.room_state != RoomState::Lobby {
-                    return;
-                }
-
-                let _ = self.tx.send(Event::Forward {
-                    inner: room_event::Event::Comment(inner),
-                });
+            tracing::info!("{:?} {:?}", client_index, race);
```
GnomedDev, line 170: Info is for useful but not important logs, this should be `debug`

tools/gameserver/src/room.rs
```diff
@@ -1,256 +1,370 @@
...
+            }
+            let player_times = player_frames.iter().map(|(time, _)| *time).collect();
+            let players =
+                player_frames.iter().map(|(_, player_frame)| player_frame.clone()).collect();
+            let server_frame = RaceServerFrame {
+                time,
+                player_times,
+                players,
+            };
+            tracing::info!("{:?}", server_frame);
```
GnomedDev, line 207: Same with above.

tools/gameserver/src/room.rs
```diff
@@ -1,256 +1,370 @@
...
-                    let _ = self.tx.send(Event::Forward {
-                        inner: room_event::Event::SelectInfo(room_event::SelectInfo {
-                            player_properties,
-                            selected_player: winning_vote as u32,
-                        }),
-                    });
+        let disconnect_tx = self.disconnect_tx.clone();
+        let read_tx = self.read_tx.clone();
+        let mut write_rx = self.write_tx.subscribe();
+        let task = tokio::spawn(async move {
```
GnomedDev, line 317: This should be moved to it's own function, avoid having to spawn an empty task just to replace it.

tools/gameserver/src/unreliable_socket.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    pub fn connection_addr(&self, index: usize) -> Option<SocketAddr> {
+        self.connections[index].addr
+    }
+
+    pub async fn read<M>(&mut self) -> Result<(usize, M)>
+    where
+        M: Message + Default,
+    {
+        let mut tmp = [0u8; 1024]; // TODO make that configurable
```
GnomedDev, line 36:
```suggestion
        let mut buf = [0u8; 1024]; // TODO make that configurable
```

tools/netprotocol/src/async_stream.rs
```diff
@@ -38,64 +47,194 @@ impl<R: Message + Default, W: Message> AsyncStream<R, W> {
...
+                *self.offset += size;
+                Ok(())
+            }
+            Err(e) => Err(e.into()),
+        }
+    }
 
-        Ok(Some(R::decode(&*msg)?))
+    pub fn start_write(&mut self, message: &W) {
+        assert!(*self.size == 0);
```
GnomedDev, line 218: This check could be made mandatory by using an `Option<NonZeroUsize>`

GnomedDev (changes requested):
Looks good, just some small concerns.

`.ok_or(anyhow!("whatever"))` should also be replaced with `.context("whatever")`, wherever it is used.

tools/gameserver/Cargo.toml
stblr: This was used for iteration in one of the prototypes but it looks like this is not needed anymore. Will swap it back because there is no convenient way to get the key before constructing the value.

tools/netprotocol/src/async_stream.rs
stblr: I will see if there is a way to implement the checks at compile time (have `start_write` return a struct with a mutable reference).

tools/gameserver/src/room.rs
stblr: That syntax doesn't look to be legal.

GnomedDev: Oh, huh, didn't actually know `tokio::select!` had that `else` syntax.

tools/gameserver/Cargo.toml
stblr: Went back with slab, will see if I can implement faster iteration and submit it upstream, or if I'm missing something.

tools/gameserver/src/unreliable_socket.rs
stblr: Renamed it to `message` like in `AsyncStream`.

tools/netprotocol/src/async_stream.rs
stblr: Dropped cancel safety for writes as well as split reader/writer, as this is no longer needed.

PR #568, opened 2023-02-05
Add eggScnRenderer headers

riidefi:


payload/egg/gfx/eggScnRenderer.hpp
```diff
@@ -0,0 +1,24 @@
...
+    DrawPathBase *getDrawPathBase(u16 path) const {
+        // assert(path < getNumDrawPath());
+        return mppPathSet[path];
+    }
+
+private:
+    // ScnRendererConfig VT at 0xB4
+    // CustomScnProc from 0xB8 to 0xC4
+    char _b4[0xc4 - 0xb4];
+    DrawPathBase **mppPathSet;
```
vabold, line 19: `m_` prefix instead of `mpp`

payload/egg/gfx/eggScnRenderer.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include "egg/gfx/eggScnRootEx.hpp"
+
+namespace EGG {
+
+class ScnRenderer : public ScnRootEx {
+public:
+    DrawPathBase *getDrawPathBase(u16 path) const {
+        // assert(path < getNumDrawPath());
+        return mppPathSet[path];
+    }
```
vabold, line 13: You could maybe define this in a `.cc` file as `DrawPathBase *ScnRenderer::drawPathBase(u16 path) const;`.

vabold (changes requested):
`.hh`

payload/egg/gfx/eggScnRenderer.hpp
riidefi: (This name comes from N)

PR #573, opened 2023-02-06
Save the character and vehicle combination that was last used

MikeIsAStar:


stblr (approved):
Nice work!

MikeIsAStar:
> Nice work!

Thank you !

PR #574, opened 2023-02-06
Split graphical settings off into their own category

GnomedDev:


riidefi:
This is an improvement so should be merged. All but FOV settings should exist in a "HUD" tab ideally.

stblr:
> This is an improvement so should be merged. All but FOV settings should exist in a "HUD" tab ideally.

I think this should be done in this PR.

riidefi:
Can we re-open and rename to HUD?

PR #578, opened 2023-02-09
Remove the rest of Server Mode

GnomedDev:
This removes the remaining `RoomServer` files, the sections for server mode, and stubs the button for Server Mode. This doesn't seem to actually shorten the payload size, but my testing was probably wrong.


stblr (approved):
Let's go!

PR #582, opened 2023-02-12
Do not permit the use of a modified 'StaticR.rel' file

MikeIsAStar:


payload/sp/Rel.cc
```diff
@@ -20,8 +20,87 @@ namespace SP::Rel {
...
+
+    u32 relSize = relModuleInfoArray[regionIndex].size;
+    if (roundedRelSize < relSize) {
+        return false;
+    }
+
+    u8 relHash[NET_SHA1_DIGEST_SIZE];
+    NETSHA1Context context;
+    NETSHA1Init(&context);
+    NETSHA1Update(&context, rel, relSize);
```
riidefi, line 98: How long does this operation take?

vabold: I watched an episode of The Simpsons and it loaded before the episode ended.

vabold:
Looks good to merge, performance tested with Simpsons S08E23 "Homer's Enemy".

stblr (approved):

PR #584, opened 2023-02-12
Make instant menus default

riidefi:


Zachruff:
no

william-texas:
no

MikeIsAStar:
LOL

vabold:
Removing animations as the default option only presents SP as unpolished, especially given that we have to wait for things like icons to load in (e.g. the vehicle select page).

PR #594, opened 2023-02-15
Cleanup configure.py

GnomedDev:
- Uses built-in `glob` module to prevent having to list every file manually
- Properly checks for `os.environ.get` returning `None` or `""
- Sorts imports according to default VScode behaviour (and removes unused `subprocess` import)
- Avoids pylance getting confused by removing use of `target` after the for loop defining it.


riidefi:
This should simplify things. Though with the globs, we shouldn't include the files in the `Emulator` file in actual builds. Implementing a few APIs, they allow some SP classes to run sans-modification on PC targets. We might want to simply remove them altogether.

stblr (changes requested):
The README should be updated to tell users to run `configure.py` every time a file is added or removed.

stblr:
Should we wait for a PyPy release that supports 3.10? PyPy is still recommended in the README, but the gains are likely more limited now with the move to LZMA.

GnomedDev:
`configure.py` doesn't even take that long, and seeing that we are running lots of small scripts I can imagine pypy isn't helping that much (no time for JIT to warm up, might even be slower thinking about it)

README.md
```diff
@@ -29,7 +29,8 @@ ninja
 
 The `out` directory will contain the generated binaries and assets.
 
-On subsequent builds only the last command needs to be run.
+The `configure.py` command only needs to be run when new files have been
+added to the project, and can be skipped if only recompiling with changed files.
```
stblr, line 33: Or removed.

stblr (changes requested):
PyPy should no longer be mentioned in the README.

stblr (approved):

riidefi:
This breaks Windows builds

vabold:
inb4 "good"

PR #595, opened 2023-02-15
Remove duplicate `"font": "regular"` fields from translation files

GnomedDev:


tools/bmg-cleaner/bmg-cleaner.py
```diff
@@ -0,0 +1,25 @@
+import json, json5
```
stblr, line 1: The first import is unused.

tools/bmg-cleaner/bmg-cleaner.py
```diff
@@ -0,0 +1,25 @@
...
+    assert isinstance(bmg, dict)    
+
+    for message_id, message in bmg.items():
+        font = message.get("font")
+
+        new_bmg[message_id] = {"string": message["string"]}
+        if font is not None and font != "regular": 
+            new_bmg[message_id]["font"] = font
+
+    path.write_text(json5.dumps(new_bmg, indent=4, ensure_ascii=False))
```
stblr, line 22: What about `quote_keys`?

stblr (changes requested):

tools/bmg-cleaner/bmg-cleaner.py
GnomedDev: What about it? If we are using json5, shouldn't we be taking advantage of it's features?

stblr: Yeah why not.

stblr (approved):

PR #596, opened 2023-02-18
Implement MVP Online UI

GnomedDev:
This PR implements the minimum viable product for online/matchmaking UI.

A couple of things to take note of:
- `OnlineConnectionManagerPage` transitions away to an error as soon as it fails to connect, meaning direct connection cannot be used without working matchmaking servers, this should be fixed before release
-  `OnlineConnectionManagerPage` has to override `dt` for the `AsyncSocket` to be correctly destroyed, this should be looked on a larger scale (are we leaking memory due to most page destructors not being called?)
- `OnlineTopPage` hasn't implemented the friends page or track packs, and currently the buttons do nothing
- ~~`OnlineModeSelectPage` only shows 5000VR, which is probably a fallback default, I'll look into where that is set soon~~ Done
- `RandomMatchingPage`'s current `RoomClient` handler may mean that the first member of a Room could send settings and have the Room respect it?


payload/game/ui/FriendButton.hh
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include "game/ui/Button.hh"
+
+namespace UI {
+
+class FriendButton : public PushButton {
+public:
+    void load(const char** animInfo, const char* dir, const char* file, const char* variant, bool r8, bool pointerOnly);
```
stblr, line 9: What about the other member functions?

configure.py
```diff
@@ -1493,6 +1494,7 @@
         os.path.join('payload', 'game', 'ui', 'RaceConfirmPage.cc'),
         os.path.join('payload', 'game', 'ui', 'RadioButtonControl.cc'),
         os.path.join('payload', 'game', 'ui', 'RandomMatchingPage.cc'),
+        os.path.join('payload', 'game', 'ui', 'OnlineConnectionManagerPage.cc'),
```
stblr, line 1497: Ordering

tools/gameserver/src/room.rs
```diff
@@ -121,10 +136,8 @@ impl Room {
                     }
                     match self.handle_select_request(client_key, request, gamemode) {
                         Ok(Some(gamemode)) => return Ok(Some(gamemode)),
-                        Err(e) => {
-                            // TODO handle
-                        },
-                        _ => (),
+                        Err(err) => tracing::error!("Failed to handle select request: {err}"),
```
stblr, line 139: The TODO comment can be kept.

payload/sp/cs/RoomClient.cc
```diff
@@ -601,13 +614,26 @@ bool RoomClient::writeVote(u32 course, std::optional<Player::Properties> propert
...
+void RoomClient::TransitionToError(u32 errorCode) {
+    auto *sectionManager = UI::SectionManager::Instance();
+    auto *globalContext = sectionManager->globalContext();
+
+    globalContext->m_onlineDisconnectInfo.m_category = UI::OnlineErrorCategory::ErrorCode;
+    globalContext->m_onlineDisconnectInfo.m_errorCode = errorCode;
+
+    sectionManager->setNextSection(UI::SectionId::OnlineDisconnected, UI::Page::Anim::None);
+    sectionManager->startChangeSection(0, 0);
 }
```
stblr, line 637: I would rather suggest to add a callback to the handler to prevent frontend details from leaking into the backend.

payload/game/ui/OnlineConnectionManagerPage.cc
```diff
@@ -0,0 +1,223 @@
...
+    m_state = State::Initial;
+};
+
+void OnlineConnectionManagerPage::dt(s32 type) {
+    if (type > 0) {
+        delete this;
+    } else {
+        this->~OnlineConnectionManagerPage();
+    }
+}
```
stblr, line 27: The base class already does this AFAIK.

payload/game/ui/OnlineTopPage.cc
```diff
@@ -2,76 +2,131 @@
...
-    m_connectButton.setFrontHandler(&m_onConnectButtonFront, false);
-    m_backButton.setFrontHandler(&m_onBackButtonFront, false);
+void OnlineTopPage::onActivate() {
+    m_replacement = PageId::None;
+    m_worldwideButton.selectDefault(0);
+    m_instructionText.setMessage(0x10d6);
 
     if (SectionManager::Instance()->currentSection()->id() == SectionId::OnlineSingle) {
-        m_pageTitleText.setMessage(20003);
+        m_pageTitleText.setMessage(0x7f1);
```
stblr, line 76: Decimal

payload/game/ui/OnlineTopPage.cc
```diff
@@ -2,76 +2,131 @@
...
     } else {
-        m_pageTitleText.setMessage(20004);
+        m_pageTitleText.setMessage(0x7f2);
```
stblr, line 78: Same

stblr (changes requested):

payload/game/ui/OnlineConnectionManagerPage.cc
GnomedDev: The base class's implementation does not end up calling the `AsyncSocket` destructor and therefore caused "Client already logged in" issues as mentioned in the PR description.

payload/game/ui/FriendButton.hh
GnomedDev: What do you mean? I implemented the bare minimum to get the button to load, as this is a MVP and other FriendButton stuff can be implemented later.

stblr: The `FriendButton` vfuncs won't be called so this is just a regular `PushButton`.

GnomedDev: Honestly probably for the best, a lot of the friend code is directly calling gamespy stuff (hence the `WifiFriendMenu` being deleted and the button being stubbed), the only reason this I even call to this load function is to avoid a broken animation.

stblr: Since this is temporary and doesn't break anything, I guess this is fine to have this hybrid class for now.

payload/sp/cs/RoomClient.cc
GnomedDev: Wouldn't that remove the ability to report `30004` (`RoomClient` missing), as a RoomClient would have to be there to call the handler and calling a method with a null `this` pointer is UB.

payload/game/ui/OnlineModeSelectPage.cc
```diff
@@ -0,0 +1,102 @@
...
+
+#include "game/ui/SectionManager.hh"
+
+namespace UI {
+
+void OnlineModeSelectPage::setRatings(u16 vsRating, u16 btRating) {
+    MessageInfo info = {};
+
+    info.intVals[0] = vsRating;
+    m_raceButton.setMessage("go", 0x106a, &info);
```
stblr, line 11: Decimal

payload/game/ui/OnlineModeSelectPage.cc
```diff
@@ -0,0 +1,102 @@
...
+
+    info.intVals[0] = btRating;
+    m_battleButton.setMessage("go", 0x106b, &info);
```
stblr, line 14: Same

payload/game/ui/RandomMatchingPage.cc
```diff
@@ -23,161 +18,71 @@ PageId RandomMatchingPage::getReplacement() {
...
-        response.message.start_matchmaking.gamemode = 0;
-
-        m_state = State::WaitForMatch;
-        break;
-    default:
-        panic("unexpected event");
+    if (onlineManager->isCustomTrackpack()) {
+        m_title.setMessage(0xfa1);
+    } else {
+        m_title.setMessage(0xfa0);
```
stblr, line 44: Decimal

payload/game/ui/OnlineModeSelectPage.cc
```diff
@@ -0,0 +1,102 @@
...
+    SP_LOG("OnlineModeSelectPage::onActivate");
+
+    auto section = SectionManager::Instance()->currentSection();
+    auto connectionManager = section->page<PageId::OnlineConnectionManager>();
+
+    if (connectionManager->isCustomTrackpack()) {
+        // TODO: Show the name of the pack selected?
+        m_titleText.setMessage(0xfa1);
+    } else {
+        m_titleText.setMessage(0xfa0);
```
stblr, line 65: Decimal

payload/game/ui/OnlineTopPage.cc
```diff
@@ -2,76 +2,131 @@
...
-void OnlineTopPage::onActivate() {
+void OnlineTopPage::onBack(u32 localPlayerId) {
     m_replacement = PageId::None;
+    push(PageId::WifiDisconnect, Anim::None);
 }
 
-void OnlineTopPage::onBack([[maybe_unused]] u32 localPlayerId) {
-    changeSection(SectionId::TitleFromMenu, Anim::Prev, 0.0f);
+void OnlineTopPage::onButtonSelect(PushButton* button, [[maybe_unused]] u32 localPlayerId) {
+    m_instructionText.setMessage(0x10d6 + button->m_index);
```
stblr, line 88: Decimal

payload/game/ui/OnlineConnectionManagerPage.cc
stblr: We should at least open an issue to investigate this.

payload/sp/cs/RoomClient.cc
stblr: You could make this a method of e.g. `SectionManager` or `GlobalContext` that would be called by the callback but also when `RoomClient` is missing.

stblr (approved):

PR #604, opened 2023-02-25
Add basic save state system

GnomedDev:
Allows users (or the mod, for a future practice mode) to save the state of a kart and restore it later, without needing heavy dolphin savestates. This technically saves a lot more than necessary, as I went through trying to find what would stop desyncs, but the headers could be useful for future use.

`KartSaveState` is currently sized at `4560` (`0x11d0`), so it isn't a massive memory hog.

This could be bound to a button, which I may do in this PR or a future PR depending on how long it takes to merge, but is currently bound to the `/store` and `/reload` commands.


configure.py
```diff
@@ -1378,6 +1378,7 @@
         os.path.join('payload', 'game', 'kart', 'KartMove.cc'),
         os.path.join('payload', 'game', 'kart', 'KartObject.S'),
         os.path.join('payload', 'game', 'kart', 'KartObject.cc'),
+        os.path.join('payload', 'game', 'kart', 'KartSaveState.cc'),
```
stblr, line 1381: Ordering

payload/egg/core/eggSystem.cc
```diff
@@ -6,6 +6,8 @@ extern "C" {
 #include <sp/Commands.h>
 #include <sp/IOSDolphin.hh>
 #include <game/ui/SectionManager.hh>
+#include <game/kart/KartSaveState.hh>
+#include <game/kart/KartObjectManager.hh>
```
stblr, line 10: Ordering

payload/game/kart/KartSettings.hh
```diff
@@ -1,14 +1,19 @@
...
 
 namespace Kart {
 
 struct KartSettings {
-    u8 _00[0x10 - 0x00];
+    u32 isBike;
+    u32 vehicle;
+    u32 character;
+    u16 wheelCount0; // Used for initialising suspension
+    u16 wheelCount1; // Used for intiialising tires
```
stblr, line 12: Could just be renamed to `susCount` and `tireCount`.

payload/game/kart/KartObjectProxy.hh
```diff
@@ -10,18 +10,34 @@ class KartMove;
...
 class KartState;
 class VehiclePhysics;
+class KartBody;
+class KartAction;
+class KartSus;
+class KartSettings;
+class Kart5c;
+class KartSub;
+class KartTire;
+class PlayerModel;
```
stblr, line 20: Ordering

payload/game/ui/page/RacePage.cc
```diff
@@ -49,9 +50,16 @@ RacePage *RacePage::Instance() {
...
+    Kart::kartSaveState = new Kart::KartSaveState;
     m_lastWatchedPlayerId = m_watchedPlayerId;
 }
 
+void RacePage::onDeinit() {
+    REPLACED(onDeinit)();
+
+    delete Kart::kartSaveState;
+}
+
```
stblr, line 62: Does this really belong here?

payload/game/kart/KartSettings.hh
```diff
@@ -1,14 +1,19 @@
...
     u8 playerId;
     u8 _11[0x3c - 0x11];
 };
+
```
stblr, line 16: No need for a newline.

payload/game/kart/KartSus.hh
```diff
@@ -0,0 +1,28 @@
+#include "KartObjectProxy.hh"
+
+namespace Kart {
+
+class KartSaveState;
+
+struct KartSusBase {
+    u8 _80[0x80 - 0x00];
+};
```
stblr, line 9: Is this an actual thing?

stblr (changes requested):

payload/game/kart/KartSus.hh
GnomedDev: Ah, this was KartPart in ghidra, I misread it.

payload/game/ui/page/RacePage.cc
GnomedDev: It was the first place I could think of that I could use the heap and is loaded only during the race, where else do you suggest?

stblr: `KartObjectManager::CreateInstance`, the only issue is that we are already replacing this function from C code.

payload/sp/SaveStateManager.cc
```diff
@@ -0,0 +1,82 @@
...
+#include <tuple>
+
+namespace SP {
+
+SaveStateManager *SaveStateManager::s_instance = nullptr;
+
+SaveStateManager::SaveStateManager() {
+    m_framesHeld = 0;
+    m_hasSaved = false;
+    m_kartSaveState = new Kart::KartSaveState;
```
stblr, line 17: This leaks memory.

stblr: By the way, do we even need a separate allocation?

GnomedDev: Thinking about it, probably not, I implemented this way to lazy allocate the save state but that doesn't seem possible or useful (see dev chat last night).

stblr (approved):

PR #612, opened 2023-03-03
Avoid constructing invalid KartSaveState

GnomedDev:
Changes `KartSaveState`'s constructor to require the arguments to create a save state, avoiding the possibility of invalid junk state to be handled. This then uses std::optional in `SaveStateManager` to defer construction until the first save state is created.

### DIFF
--- a/payload/game/kart/KartSaveState.hh
+++ b/payload/game/kart/KartSaveState.hh
@@ -32,7 +32,8 @@ struct KartTireFlat {
 
 class KartSaveState {
 public:
-    KartSaveState() = default;
+    KartSaveState() = delete;
+    KartSaveState(KartAccessor accessor, VehiclePhysics* physics);
 
     void save(KartAccessor accessor, VehiclePhysics* physics);
     void reload(KartAccessor accessor, VehiclePhysics* physics);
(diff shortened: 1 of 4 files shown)


payload/game/kart/KartSaveState.hh
```diff
@@ -32,7 +32,8 @@ struct KartTireFlat {
 
 class KartSaveState {
 public:
-    KartSaveState() = default;
+    KartSaveState() = delete;
```
stblr, line 35:
```suggestion
```
That's not necessary as the default constructor is only implicitly declared if there is no other constructor.

payload/sp/SaveStateManager.cc
```diff
@@ -39,18 +39,20 @@ auto SaveStateManager::GetKartState() {
 
 void SaveStateManager::save() {
     auto [accessor, physics] = GetKartState();
-    m_kartSaveState.save(accessor, physics);
-    m_hasSaved = true;
+    if (m_kartSaveState.has_value()) {
+        (*m_kartSaveState).save(accessor, physics);
+    } else {
+        m_kartSaveState.emplace(accessor, physics);
+    }
```
stblr, line 46:
```suggestion
    m_kartSaveState.emplace(accessor, physics);
```

stblr (changes requested):

stblr (approved):

PR #613, opened 2023-03-05
Allow users to return to the Wii System Menu from the exception handler

MikeIsAStar:


vabold:
Jos in the Dolphin Discord server mentioned that we could try using "whatever homebrew programs do to return to the loader", and since Dolphin doesn't have a loader it would close emulation.

We could try experimenting with this if we want, or we could instead relegate it to another PR. I think relegating it to another PR is fine since we take proper precautions to disallow Dolphin from accessing this.

vabold:
After some testing, we've concluded that the solution is to PR another ioctl to Dolphin. As it stands, this PR is good for review.

vabold (approved):

stblr:
> After some testing, we've concluded that the solution is to PR another ioctl to Dolphin. As it stands, this PR is good for review.

This looks more like a workaround.

stblr (approved):

PR #615, opened 2023-03-06
Fix Windows builds

GnomedDev:
For some god awful reason, this works on Linux and not Windows. Due to the lack of Windows CI, this caused #594 to break Windows builds (not that anyone should care).


stblr:
It looks like too many files were included.

GnomedDev:
Yeah, just why this worked on linux but not windows is a mystery

Issue #530, opened 2023-01-08
OOM in Change Ghost Data


stblr:

GnomedDev:
Possibly related to #606

stblr:
Possibly fixed by #620.

PR #620, opened 2023-03-06
Remove the course cache

stblr:
This seems to work but needs more testing as this has the potential to break a lot of things.


nicl83:
Was asked by @GnomedDev to try a build from this branch after hitting an OoM when trying to change ghost data in a Time Trial. Seems to fix #530 and haven't noticed any major issues, but haven't tried local or online multiplayer yet

stblr:
> Was asked by @GnomedDev to try a build from this branch after hitting an OoM when trying to change ghost data in a Time Trial. Seems to fix #530 and haven't noticed any major issues, but haven't tried local or online multiplayer yet

Thanks for the feedback!

stblr:
For this we still need to think about track identification.

GnomedDev:
What do you mean by that?

stblr:
With the course cache we could easily get a hash of the track to filter ghosts (CTGP and MKW-SP ghosts store that in the ghost file). The simplest way to work around that would be to hash everything on startup but that's way too slow and doesn't handle NetStorage.

PR #628, opened 2023-03-07
Implement consistent style for declaring unused variables

GnomedDev:
This PR removes usage of `maybe_unused` and the `unused` macro and replaces with `/* arg */`. This avoids the ambiguity of maybe_unused and helps codebase uniformity

### DIFF
--- a/include/Common.h
+++ b/include/Common.h
@@ -66,6 +66,7 @@ typedef double f64;
         _a > _b ? _a : _b;      \
     })
 
+// Only for use in vendor code, we use `/* ... */` comments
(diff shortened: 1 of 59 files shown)


include/Common.h
```diff
@@ -66,6 +66,7 @@ typedef double f64;
         _a > _b ? _a : _b;      \
     })
 
+// Only for use in vendor code, we use `/* ... */` comments
```
stblr, line 69: If by vendor code you mean `diskio.cc`, this is actually our code (that we have to provide to the library).

stblr (approved):

PR #635, opened 2023-03-15
Make debug builds optional

GnomedDev:
This PR adds a `--default-targets` argument to the `configure.py` script, which defaults to only build the `test` target. This speeds up compile time by 2x due to having to only compile the `test` target, and avoiding the `debug` target which currently is too large to launch on console.

### DIFF
--- a/.github/workflows/build_sp.yml
+++ b/.github/workflows/build_sp.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Install protobuf
         run: pip install protobuf
       - name: Configure ninja
-        run: ./configure.py --ci
+        run: ./configure.py --ci --default-targets test,debug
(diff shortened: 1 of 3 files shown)


stblr:
I'd rather fix our memory usage here, and if we want to speed up the CI we can run both in parallel.

GnomedDev:
I meant this for personal builds, as I switch branches often, swap between gdb and non-gdb compat for debug and builds for friends, and use multiple machines, it seems a waste to compile a target I never use and thought it would be a waste for others aswell.

stblr:
Then the CI should override the default targets to keep the debug build.

GnomedDev:
Done

stblr (approved):

PR #652, opened 2023-03-31
SaveStates: Fix double save causing crash

GnomedDev:
Turns out calling the destructor of KartSaveState is really bad, due to the somewhat jank nature of it.


GnomedDev:
Closing in favor of #729

PR #655, opened 2023-04-01
Implement basic Track Pack System

GnomedDev:
This implements the Track Pack system somewhat described in the document and allows for pack authors to bundle custom tracks with a manifest to allow users to play CTs without needing the jank of "My Stuff".

This system will read `.ini` files from the `mkw-sp/Track Packs` folder, which contains metadata and lists of tracks for each mode.
Tracks are stored in the `mkw-sp/Tracks` folder, which stores tracks as `sha1.ext` (where ext is `.szs` or `.arc.lzma`) and this `sha1` is looked up in `mkw-sp/TrackDB.ini` to fetch name and slot information. Thumbnails are also looked up using this `sha1` in `mkw-sp/Track Thumbnails/sha1.jpg`.

Track developers can setup a "development pack" using their track's sha1, however this may be tedious (having to update the name and the pack file for every new revision) 

**Regressions caused**:
- Online UI no longer loads, due to missing files for pack select

**Known Issues**:
- TrackPack::getTrackList panics if TrackGameMode isn't exactly a specific mode, instead of returning a merged vector.


payload/egg/core/eggTaskThread.hh
```diff
@@ -4,11 +4,12 @@
...
 namespace EGG {
 
+template <typename T>
 class TaskThread {
 public:
-    typedef void (*TFunction)(void *);
+    typedef void (*TFunction)(T);
 
-    void request(TFunction mainFunction, void *arg, void *taskEndMessage);
+    void request(TFunction mainFunction, T arg, void *taskEndMessage);
```
riidefi, line 12: This change doesn't make any sense. Taking an arbitrary `void*` as user data is standard practice in C code. `TaskThread` isn't valid for any arbitrary type of `arg`. If you want a safe interface, make a wrapper type or freestanding function.

payload/sp/TrackPackManager.cc
```diff
@@ -0,0 +1,391 @@
...
+    u32 out = 0;
+
+    auto first = sv.data();
+    auto last = sv.data() + sv.size();
+    auto [ptr, err] = std::from_chars(first, last, out);
+
+    if (ptr == last) {
+        return out;
+    } else {
+        panic("Failed to parse integer!");
```
riidefi, line 97: Ideally we wouldn't crash from an invalid .ini file. Not a blocker but perhaps we could adapt this function to return `std::expected<u32, FixedString<32>>` or such?

payload/sp/TrackPackManager.cc
```diff
@@ -0,0 +1,391 @@
...
+    return parsedTrackIds;
+}
+
+TrackPack::TrackPack(std::string_view manifestView) {
+    IniReader iniReader(manifestView);
+
+    bool prettyNameFound = false;
+    bool descriptionFound = false;
+    bool authorNamesFound = false;
+    while (auto property = iniReader.next()) {
```
riidefi, line 133: auto [section, key] = iniReader.next()

payload/sp/TrackPackManager.cc
```diff
@@ -0,0 +1,391 @@
...
+
+void TrackPackInfo::selectCourse(u32 wiimmId, TrackGameMode mode) {
+    auto &trackPackManager = TrackPackManager::Instance();
+
+    m_selectedWiimmId = wiimmId;
+    m_selectedCourseId = trackPackManager.getCourseId(wiimmId, mode);
+}
+
+TrackGameMode getTrackGameMode(u32 gameMode, u32 battleType) {
+    if (gameMode == 1 || gameMode == 2) {
```
riidefi, line 376: Do we not have an enum for this?

payload/sp/vanillaTracks.hh
```diff
@@ -0,0 +1,9 @@
+const std::string_view vanillaManifest = R"(
```
riidefi, line 1: I believe this is invalid if included in multiple files. You can fix that by making it `constexpr`, which may also allow for some additional optimization.

payload/sp/TrackPackManager.cc
GnomedDev: What would the user facing behavior you would like to see? If one of the pack files is invalid for now, and we didn't panic, it would need some UI code to implement a dialog or something, which should probably be done in a follow up PR.

payload/egg/core/eggTaskThread.hh
GnomedDev: Ah yes, I did this while debugging issues and never reverted it, I'll revert now.

payload/sp/vanillaTracks.hh
riidefi: If you are not allowed to `#include` the file, then it is not a header.

GnomedDev: It shouldn't be a header anyway, this is just making up for the fact C++ doesn't have a form of include_str.

payload/sp/TrackPackManager.cc
GnomedDev: This doesn't seem to compile?

riidefi: `*property`*

payload/sp/vanillaTracks.hh
riidefi: Probably better to make it either `constexpr` or `inline` so it's a valid header for now. If we want to avoid duplicate bodies being generated we can put it in its own .cc file later, but I think the linker will merge them anyway since it's const.

payload/sp/TrackPackManager.cc
riidefi: Fair

payload/sp/FixedString.hh
```diff
@@ -41,4 +47,27 @@ struct FixedString {
...
+struct WFixedString {
+    constexpr WFixedString() = default;
+    constexpr WFixedString(const WFixedString &) = default;
+    constexpr ~WFixedString() = default;
+    constexpr WFixedString(std::string_view view) {
+        if (view.size() >= m_buf.size()) {
+            view = view.substr(0, N - 1);
+        }
+
+        auto written = Util_toUtf16(m_buf.data(), m_buf.size(), view.data(), view.size());
```
riidefi, line 60: m_len = ...

payload/game/ui/GhostManagerPage.cc
```diff
@@ -16,18 +16,34 @@ void GhostManagerPage::SPList::populate() {
...
         }
+
         if (header->courseId != courseId) {
             continue;
+        };
+
+        auto sha1 = footer->courseSHA1();
+        if (!sha1.has_value()) {
+            continue;
+        };
```
riidefi, line 38: Stray semicolons?

payload/game/ui/Section.cc
```diff
@@ -191,13 +192,15 @@ void Section::addPage(PageId pageId) {
         }
     }
 
-    SP_LOG("addPage %u", pageId);
+    SP_LOG("addPage 0x%x", pageId);
```
riidefi, line 195: Print both hex & decimal ideally

payload/game/ui/Section.cc
```diff
@@ -222,7 +225,7 @@ void Section::addActivePage(PageId pageId) {
         }
     }
 
-    SP_LOG("addActivePage %u", pageId);
+    SP_LOG("addActivePage 0x%x", pageId);
```
riidefi, line 228: Ditto

payload/sp/FixedString.hh
```diff
@@ -41,4 +47,27 @@ struct FixedString {
...
+    constexpr ~WFixedString() = default;
+    constexpr WFixedString(std::string_view view) {
+        if (view.size() >= m_buf.size()) {
+            view = view.substr(0, N - 1);
+        }
+
+        m_len = Util_toUtf16(m_buf.data(), m_buf.size(), view.data(), view.size());
+        m_buf[m_len] = L'\0';
+    }
+
```
riidefi, line 63: Optional: Add constructor from std::wstring_view and conversion to std::wstring_view (to match API of FixedString)

payload/sp/ShaUtil.cc
```diff
@@ -0,0 +1,14 @@
+#include "ShaUtil.hh"
+
+#include <cstdio>
+
+std::array<char, 0x28 + 1> sha1ToHex(const Sha1 &sha1) {
```
riidefi, line 5: Probably better to return a FixedString here?

payload/sp/TrackPackManager.cc
```diff
@@ -0,0 +1,517 @@
...
+    auto [ptr, err] = std::from_chars(first, last, out);
+
+    if (ptr == last) {
+        return out;
+    } else {
+        return std::unexpected("Failed to parse integer!");
+    }
+}
+
+std::expected<Sha1, const char *> sha1FromSv(std::string_view sv) {
```
riidefi, line 94: Move to ShaUtil?

payload/sp/TrackPackManager.cc
```diff
@@ -0,0 +1,517 @@
...
+    std::vector<Sha1> parsedTrackIds;
+    char buf[(sizeof(Sha1) * 2) + 1];
+    size_t position = 0;
+
+    for (char c : tracks) {
+        if (c == ',' || position == sizeof(buf)) {
+            buf[position] = '\0';
+
+            std::string_view sv(buf, position);
+            auto trackId = sha1FromSv(sv);
```
riidefi, line 150: TRY should clean this up a bit

payload/sp/TrackPackManager.cc
```diff
@@ -0,0 +1,517 @@
...
+    return supportedModes;
+}
+
+u16 TrackPack::getTrackCount(TrackGameMode mode) const {
+    assert(m_parseError == nullptr);
+    return getTrackList(mode).size();
+}
+
+std::optional<Sha1> TrackPack::getNthTrack(u32 n, TrackGameMode mode) const {
+    assert(m_parseError == nullptr);
```
riidefi, line 292: See this is why I'm not a huge fan of constructors in C++ turning every type into an optional type.

riidefi (changes requested):

payload/sp/ShaUtil.cc
GnomedDev: Doesn't FixedString have capacity and length? Because it will always be the same amount of chars, this is simpler imo.

payload/sp/TrackPackManager.cc
```diff
@@ -0,0 +1,470 @@
...
+
+std::expected<std::vector<Sha1>, const char *> parseTracks(std::string_view tracks) {
+    std::vector<Sha1> parsedTrackIds;
+    char buf[(sizeof(Sha1) * 2) + 1];
+    size_t position = 0;
+
+    for (char c : tracks) {
+        if (c == ',' || position == sizeof(buf)) {
+            buf[position] = '\0';
+            std::string_view sv(buf, position);
```
riidefi, line 127: `std::string_view` need not be null-terminated, so we should be able to avoid a temporary buffer (like `buf`) entirely.

payload/sp/TrackPackManager.cc
```diff
@@ -0,0 +1,470 @@
...
+            } else if (key == "balloon") {
+                m_balloonTracks = TRY(parseTracks(value));
+            } else if (key == "coin") {
+                m_coinTracks = TRY(parseTracks(value));
+            } else {
+                return std::unexpected("Unknown key in track pack manifest");
+            }
+        } else {
+            auto sha1 = TRY(sha1FromSv(section));
+            if (m_unreleasedTracks.size() == 0 || m_unreleasedTracks.back().sha1 != sha1) {
```
riidefi, line 181: Usually more idiomatic to use `.empty()` instead of `.size() == 0`

payload/sp/TrackPackManager.cc
```diff
@@ -0,0 +1,470 @@
...
+    while (auto property = trackDbIni.next()) {
+        auto [section, key, value] = *property;
+
+        auto sha1 = sha1FromSv(section);
+        if (!sha1) {
+            SP_LOG("Could not parse sha1!");
+            continue;
+        }
+
+        if (m_trackDb.size() == 0 || m_trackDb.back().sha1 != *sha1) {
```
riidefi, line 336: ditto

payload/sp/TrackPackManager.cc
```diff
@@ -0,0 +1,470 @@
...
+    return m_packs[trackPackInfo.m_selectedTrackPack];
+}
+
+void TrackPackInfo::setTrackMessage(UI::LayoutUIControl *control) const {
+    return setTrackMessage(control, s_name.c_str(), m_selectedCourseId);
+}
+
+void TrackPackInfo::setTrackMessage(UI::LayoutUIControl *control, const wchar_t *name,
+        u32 courseId) const {
+    auto raceConfig = System::RaceConfig::Instance();
```
riidefi, line 366: auto*

payload/sp/TrackPackManager.hh
```diff
@@ -0,0 +1,123 @@
...
+};
+
+class Track {
+public:
+    Track(Sha1 sha1) : sha1(sha1){};
+
+    std::expected<void, const char *> parse(std::string_view key, std::string_view value);
+    u32 getCourseId() const;
+
+    WFixedString<64> name = {};
```
riidefi, line 35: Probably makes sense to put `name` last for cache reasons?

payload/sp/TrackPackManager.hh
```diff
@@ -0,0 +1,123 @@
...
+    u32 m_selectedTrackPack = 0;
+
+private:
+    // Private as need to be kept in sync
+    u32 m_selectedCourseId = 0;
+    u32 m_selectedMusicId = 0;
+    Sha1 m_selectedSha1 = {};
+
+    // We don't have enough space to store this otherwise.
+    static WFixedString<64> s_name;
```
riidefi, line 120: std::unique_ptr<WFixedString<64>> maybe?

payload/sp/ShaUtil.cc
```diff
@@ -0,0 +1,34 @@
...
+
+#include <cstdio>
+#include <cstdlib>
+
+std::array<char, 0x28 + 1> sha1ToHex(const Sha1 &sha1) {
+    std::array<char, 0x28 + 1> hex;
+
+    const u8 *data = sha1.data();
+    for (u32 i = 0; i < sha1.size(); ++i) {
+        sprintf(hex.data() + (i * 2), "%02x", (*(data + i)) & 0xff);
```
riidefi, line 11: Can you adapt to use `snprintf`? Wouldn't mind completely blacklisting this function

riidefi (changes requested):

payload/sp/ShaUtil.cc
riidefi: That's true but it seems unintuitive to see `.data()` be valid for a cstr later. If this isn't performance critical, it'd probably be a lot more clear to suffer the 4 byte penalty and use a `FixedString`.

payload/sp/TrackPackManager.cc
```diff
@@ -0,0 +1,470 @@
...
+        return std::unexpected("Missing required key in track pack manifest");
+    } else if (m_raceTracks.empty() && m_balloonTracks.empty() && m_coinTracks.empty()) {
+        return std::unexpected("No tracks found in track pack manifest");
+    }
+
+    return {};
+}
+
+const Track *TrackPack::getUnreleasedTrack(Sha1 sha1) const {
+    for (const auto &track : m_unreleasedTracks) {
```
riidefi, line 200: If you `std::sort` this array earlier, you can use `std::lower_bound` to perform a binary search

riidefi: (optional)

riidefi: Probably just slower here since you reload the ini files every race (?)

GnomedDev: I reload every time the pack selection page is initialised, or every race when capturing thumbnails

payload/sp/TrackPackManager.hh
GnomedDev: Wouldn't `unique_ptr` heap allocate the `WFixedString`? That wouldn't work as we need to keep all this data across the scene boundary (heap cleared).

riidefi: I suppose you could allocate in the System heap. It is probably better to just extend the structure and place TrackPackManager at the end (allocate more memory).

GnomedDev: The system heap stays allocated? Could I put `TrackPackManager` in there and get rid of `TrackPackInfo` entirely?

riidefi: It's better to keep the data in MEM2 and extend the structure you insert TrackPackManager into

payload/sp/TrackPackManager.hh
GnomedDev: Hmm? That seems like a micro optimisation and even then, how would that be better for cache reason?

payload/sp/FixedString.hh
```diff
@@ -41,4 +47,27 @@ struct FixedString {
     std::array<char, N> m_buf = {};
 };
 
+template <size_t N>
+struct WFixedString {
+    constexpr WFixedString() = default;
+    constexpr WFixedString(const WFixedString &) = default;
+    constexpr ~WFixedString() = default;
+    constexpr WFixedString(std::string_view view) {
```
riidefi, line 55: Probably better as a static function like `fromUTF8`, possibly returning an optional value down the road?

payload/sp/vanillaTracks.hh
```diff
@@ -0,0 +1,261 @@
+constexpr std::string_view vanillaManifest = R"(
```
riidefi, line 1: `#pragma once`

payload/game/system/MultiDvdArchive.hh
```diff
@@ -20,19 +20,23 @@ public:
 
     MultiDvdArchive(u16 archiveCount);
     virtual ~MultiDvdArchive();
+    virtual void dt(s32 type);
+
+    REPLACE virtual void init();
+    void REPLACED(init)();
```
riidefi, line 26: Maybe also pull this into its own PR as a bugfix?

payload/game/ui/SingleTopPage.cc
```diff
@@ -105,10 +110,38 @@ void SingleTopPage::onActivate() {
     for (u32 i = 0; i < 12; i++) {
         menuScenario.players[i].team = 2;
     }
+
+    auto &trackPack = SP::TrackPackManager::Instance().getSelectedTrackPack();
+    auto supportedModes = trackPack.getSupportedModes();
+
+    auto raceEnabled = (supportedModes & SP::TrackGameMode::Race) != SP::TrackGameMode::None;
+    auto coinEnabled = (supportedModes & SP::TrackGameMode::Coin) != SP::TrackGameMode::None;
```
riidefi, line 118: `== SP::TrackGameMode::Coin` / similar for all of these is preferable as it handles flagsets too and it's not obvious `::None` is 0.

riidefi (changes requested):

payload/sp/FixedString.hh
```diff
@@ -1,8 +1,14 @@
...
 
+extern "C" {
+#include "sp/WideUtil.h"
+
 #include <Common.h>
+}
+
 #include <algorithm>
 #include <array>
+#include <string>
```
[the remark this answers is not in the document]
GnomedDev: I have already messed this branch up with the first couple of commits, I don't want to push git any further to avoid having to merge it down into one commit.

payload/sp/FixedString.hh
```diff
@@ -41,4 +47,27 @@ struct FixedString {
     std::array<char, N> m_buf = {};
```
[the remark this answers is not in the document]
GnomedDev: Not in scope for this PR

payload/sp/vanillaTracks.hh
GnomedDev: I thought you said constexpr allows it be included multiple times?

payload/game/system/MultiDvdArchive.hh
GnomedDev: It isn't a bugfix to upstream, as it isn't used in upstream. Just seems like mindless busywork.

payload/sp/vanillaTracks.hh
riidefi: All other header files in the repo begin with `#pragma once#. This one isn't special.

`constexpr` makes the definition weak which prevents linker ODR-violations if the file is included in multiple translation units. It's still not valid to multiply define the same weak symbol within singular translation unit.

payload/sp/FixedString.hh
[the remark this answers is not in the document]
riidefi: We should clearly communicate our maximum size for fields in the `.ini`. In this case, when you wrote `FixedString<64>`, perhaps you meant `FixedString<65>` given this behavior?

payload/sp/FixedString.hh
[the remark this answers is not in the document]
riidefi: This utility is useful beyond this PR (which cannot get merged for a while).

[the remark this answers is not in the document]
GnomedDev: Okay?

payload/game/ui/SingleTopPage.cc
GnomedDev: Just tested, and in fact, `7 == 4` does not return True!

payload/sp/FixedString.hh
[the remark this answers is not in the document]
riidefi: Gotcha. We should then enforce a limit of 63/whatever in the python wiimmdb parser.

payload/game/ui/SingleTopPage.cc
riidefi: > Just tested, and in fact, 7 == 4 does not return True!
Knock it off with the attitude. A flagset denotes a value like 0b0110 where ((0b0110 & 0b0110) != 0) == ((0b0110 & 0b0100) != 0) == ((0b0110 & 0b0010) != 0).

riidefi: Here is some testing:
```py
>>> bool(7 & 4)
True
>>> bool(7 & 2)
True
>>> bool(7 & 1)
True
```

GnomedDev: Isn't that literally it is doing though? 
> `(supportedModes & SP::TrackGameMode::Race)`

riidefi: `(supportedModes & X) == X` is preferable here since for some `X = (1 << N) | (1 << M`), `N != M`, it only triggers when both bits intersect. `(supportedModes & X) != 0` triggers when either bit intersects. So generally in C++ when we are asking if a full condition is met in a bitfield, we almost always go with the former form.

payload/game/kart/KartTire.hh
```diff
@@ -1,3 +1,5 @@
+#pragma once
```
riidefi, line 1: Would pull this into its own PR like "Added missing #pragma once". Or if possible, can we not enforce this in clang-format?

riidefi: (Since it touches a bunch of unrelated files)

GnomedDev: I did this because you were being overly pedantic about it in the glorified ini file, pull into your own PR if you want but the script I made to find them is long gone.

GnomedDev:
Closing this PR as it is no longer relevant and unmergable due to conflicts and large diff. These changes will be rewritten into smaller PRs (805, 807, 832).

PR #668, opened 2023-04-03
Remove ResFile check causing false positive bugcheck

GnomedDev:
This check was causing https://ct.wiimm.de/i/5113 to fail to load on SP, even though it loads completely fine with the check removed.

### DIFF
--- a/payload/nw4r/g3d/g3dResFile.S
+++ b/payload/nw4r/g3d/g3dResFile.S
@@ -8,16 +8,3 @@ PATCH_B_START(ResFile_CheckRevision, 0x74c)
     b InvalidRevisionFail
     // Unreachable
 PATCH_B_END(ResFile_CheckRevision, 0x74c)
-
-PATCH_B_START(ResFile_CheckRevision, 0x948)
-    cmpwi r3, 1
-    beqlr // If it's valid, do nothing
-
-    // Failsafe for unlikely revision failures
-    li r3, -1
-    li r4, 0
-    li r5, -1
-    li r6, 0
-    b InvalidRevisionFail
-    // Unreachable
-PATCH_B_END(ResFile_CheckRevision, 0x948)


riidefi:
You can add a check like this
```arm
lis r3, g_skipThisCheck@ha
lbz r3, g_skipThisCheck@l(r3)
cmpwi r3, 0
bneqlr
```
```c
bool g_skipThisCheck;
```

riidefi:
The semantics would be "allow the game to load obviously malformed and corrupted data" and the default would be `false`.

riidefi:
Keep in mind CTGP goes further than us, rejecting any track with say the `MDL0` magic missing.

GnomedDev:
Maybe spending time developing a patch to stop the game doing that but allowing the track to still load would be beneficial.

riidefi:
What, with like a neural network trained to attempt to recover useful semantic information from nondeterministically malformed source data?

riidefi:
Does GPT4 perform well on hexdumps?

GnomedDev:
Knock it off with the attitude. I mean to avoid the invalid data causing any issues (substitute default values, act like a valid input was given, etc).

riidefi:
The issue is at this stage we're probably too late. We could do some heuristics on the SZS file and attempt to detect bad files and then omit them from the game's view, though. It would be a lot of effort for a single track, and not guaranteed to work on any other track given each track can uniquely corrupt their data. Therefore simply disabling the check for this particular track seems like a reasonable middle ground.

riidefi:
I thought about this a bit more, and think we should merge this. Although we catch a bunch of malformed files, the error message we display to users is fairly unhelpful in this specific case. Really, we should scan the course, rather than trying to detect the crash as its happening.

riidefi (approved):

stblr (approved):

stblr:
Please rebase.

PR #669, opened 2023-04-04
Add rust-like TRY macro

riidefi:


GnomedDev (approved):
Looks very useful

InusualZ:
It seems that you can avoid the copy, look at the Serenity's OS [implementation](https://github.com/SerenityOS/serenity/blob/master/AK/Try.h) of the TRY & MUST macro

riidefi:
> It seems that you can avoid the copy, look at the Serenity's OS [implementation](https://github.com/SerenityOS/serenity/blob/master/AK/Try.h) of the TRY & MUST macro

Indeed, casting the result to an rvalue-reference appears to elide the copy when possible while still supporting copy-only types,
https://godbolt.org/z/nT4jrjoE8

stblr (approved):

PR #670, opened 2023-04-06
Fix text scrolling

MikeIsAStar:


stblr:
I did not test it but are you sure this is scrolling? It rather looks like the scrolling loop would not longer be reachable.

The interlaced mode changes should at the very least be moved to a separate commit.

common/Console.cc
```diff
@@ -14,19 +14,23 @@ static u8 row;
...
     cols = VI::GetXFBWidth() / Font::GetGlyphWidth() - 1;
-    rows = VI::GetXFBHeight() / Font::GetGlyphHeight() - 1;
+    rows = VI::GetXFBHeight() / Font::GetGlyphHeight() / (!VI::IsProgressive() + 1) - 1;
     col = 0;
     row = 0;
 }
 
 static void Print(char c) {
-    if (col >= cols) {
+    if (col > cols && c != '\n' && c != '\r') {
```
stblr, line 23:
```suggestion
    if (col >= cols) {
```

common/Console.cc
```diff
@@ -14,19 +14,23 @@ static u8 row;
...
 
     if (c == '\n') {
         col = 0;
-        row++;
+
+        if (row++ == rows) {
+            row = 0;
+        }
+
         return;
```
stblr, line 34:
```suggestion
        row++;

        return;
```
This break things e.g. if `Console::Print("\n\n")` is called while the console is already full. I don't think that was your intent here.

stblr (changes requested):

common/Console.cc
MikeIsAStar: Corrected in commit [6ff7b0b](https://github.com/stblr/mkw-sp/pull/670/commits/6ff7b0b67a312fa1966f90a93c6c7aa21ab251d5).

common/Console.cc
MikeIsAStar: Corrected in commit [6ff7b0b](https://github.com/stblr/mkw-sp/pull/670/commits/6ff7b0b67a312fa1966f90a93c6c7aa21ab251d5).

stblr (approved):

PR #673, opened 2023-04-09
Cleanup commands

GnomedDev:
This untangles the commands from the EGG library, and rewrites them to C++. Should be reviewable commit by commit pretty easily.

### DIFF
--- a/payload/sp/Commands.cc
+++ b/payload/sp/Commands.cc
@@ -8,6 +8,10 @@ extern "C" {
 
 #include <cstring>
 
+sp_define_command("/example_command", "Example command", const char *tmp) {
+    (void)tmp;
+}
+
 sp_define_command("/ui_info", "Dump information about the UI state to the console",
         const char *tmp) {
     (void)tmp;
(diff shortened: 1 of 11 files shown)


payload/sp/Commands.cc
```diff
@@ -0,0 +1,119 @@
...
+        return;
+    }
+
+    // Default to license 0
+    saveManager->selectSPLicense(0);
+    // TODO create base license
+    saveManager->selectLicense(0);
+
+    auto sectionId = static_cast<UI::SectionId>(nextSectionId);
+    auto anim = nextSectionId < 0 ? UI::Page::Anim::Prev : UI::Page::Anim::Next;
```
riidefi, line 71: If `nextSectionId` is valid below 0, then shouldn't the initial cast be on `abs(nextSectionId)`?

payload/sp/Commands.cc
```diff
@@ -0,0 +1,119 @@
...
+        saveStateManager->save();
+    } else {
+        OSReport("SaveStateManager not initialized\n");
+    }
+}
+
+sp_define_command("/reload", "", const char *tmp) {
+    (void)tmp;
+
+    if (auto saveStateManager = SP::SaveStateManager::Instance()) {
```
riidefi, line 37: `auto*` in all these cases otherwise good to merge

stblr (approved):

Issue #662, opened 2023-04-03
Settings do not apply.


Zachruff:
https://www.youtube.com/watch?v=Ej6ZzY6iWss 4:32 race count setting changed but vs is still 3 races.

Zachruff:
Thought it was because I accessed the settings menu from the team confirmation screen, but I tried another battle with accessing the settings from the VS/Battle/TT screen and it still didnt change behaviour (but when backing out and then checking the settings in the licence menu, the settings were the new values i had set them too despite my prior battle behaving like my old settings)

GnomedDev:
Cannot reproduce, started game -> single top page -> changed battle count to 1 -> went through normal battle UI -> battle ended after first round.

Zachruff:
Am doing more testing, It might have been from the Coin Runners/Balloon Battle page instead and im dumb.

Zachruff:
Ah ok I think I found out the thingy. If you access the settings menu anytime past the VS/TT/Battle screen, any settings you change wont apply **UNLESS** you back out to the VS/TT/Battle screen and then immediately go back in.

GnomedDev:
That makes more sense, this also applies to VS mode when team size is set to anything other than FFA, changing settings in the team confirm page will not apply until `SingleTopPage` is gone through again. 

A potential solution would be setting up all `MenuScenario` state in `CourseSelectPage`, but then changing settings in the ghost list page would desync again. The most sane option in my mind is to remove all settings buttons past `SingleTopPage`, and restrict the settings selectable while in a race.

Zachruff:
Agree with the latter, and also implement my system wide settings simplification I talked about a while ago with the flowchart. (i.e. not being able to access online settings while in vs race, etc...)

GnomedDev:
Discussed in the development channel. Conclusion reached was that settings should be reapplied during `SettingsPage::onBack` calling a shared function with `SingleTopPage`, except in a race in which the settings changes should be deferred until the race is restarted.

PR #675, opened 2023-04-11
Rewrite KartObjectManager in C++

GnomedDev:
Allows us to get rid of a lot of C bindings for this stuff


payload/game/kart/KartObjectManager.cc
```diff
@@ -1,9 +1,149 @@
...
+    switch (raceScenario.gameMode) {
+    case System::RaceConfig::GameMode::OfflineVS:
+        speedModIsEnabled = vsClass == SP::ClientSettings::EngineClass::CC200;
+        break;
+    case System::RaceConfig::GameMode::TimeAttack:
+        speedModIsEnabled = taClass == SP::ClientSettings::TAClass::CC200;
+        break;
+    default:
+        speedModIsEnabled = false;
+    }
```
stblr, line 119: Does this mean that the setting can be changed between races?

GnomedDev: Yes, just tested and it does.

stblr: That's a bug then.

stblr (changes requested):

payload/game/kart/KartObjectManager.cc
GnomedDev: How come? Sounds like a feature?

stblr: That's not consistent with the other engine classes.

GnomedDev: Those other classes are just bugged due to #662, this is moving towards the right behavior (even if it is inconsistent in the mean time)

GnomedDev: Intended behavior is that changing settings during a race does not change them until the single top page is passed through again? Do you want me to open issues for the settings that don't match this behavior?

stblr: All VS and battle settings already match that behavior AFAIK.

GnomedDev: Okay, I'll "fix" this and look for somewhere to document it.

stblr: This is just how the base game works...

stblr: Conceptually all the races in a VS game are supposed to have the same rules (with e.g. the engine class being shown in the awards). The only issue with what we have now is that it is confusing to show the VS settings in race because they will only be applied for the next game.

payload/game/system/RaceConfig.hh
```diff
@@ -120,7 +124,8 @@ public:
 private:
     REPLACE static void ConfigurePlayers(Scenario &scenario, u32 screenCount);
 
-    u8 _0000[0x0020 - 0x0000];
+    bool m_is200cc;
+    u8 _0007[0x0020 - 0x0007];
```
stblr, line 128: I would maybe put it in `MenuScenario` with the other flags. I'm also not sure where this offset change comes from.

GnomedDev: The first 0x4 bytes is the vtable, which although it doesn't get called, I didn't want to randomly overwrite it. Putting it in `m_menuScenario` would work but then it looks like it's duplicated throughout the different Scenarios, which isn't true as it is only touched in the one (eg: perfect for just being part of `RaceConfig`)

stblr: It's more for consistency than anything.

vendor/sha1/sha1.c
```diff
@@ -31,7 +31,7 @@ A million repetitions of "a"
 /* blk0() and blk() perform the initial expand. */
 /* I got the idea of expanding during the round function from SSLeay */
 #if BYTE_ORDER == LITTLE_ENDIAN
-2ewqs
+
```
riidefi, line 34: Move to its own PR

riidefi (changes requested):

vendor/sha1/sha1.c
GnomedDev: Cope

vabold: > Cope

Knock it off with the attitude.

payload/game/system/RaceConfig.hh
riidefi: Probably more readable to have the flags in one place? Jury may still be out on the proper place for them, but that seems beyond the scope of this PR.

GnomedDev: What do you mean "the flags"?

stblr (approved):

PR #680, opened 2023-04-11
Rewrite License.c to C++

GnomedDev:
This looks a bit janky, but I'm not sure how else the signature should be typed.


payload/game/ui/LicenseControl.hh
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include "UIControl.hh"
```
stblr, line 3:
```suggestion
#include "game/ui/UIControl.hh"
```

stblr (changes requested):
I think that's a returned struct.

stblr (approved):

PR #690, opened 2023-04-22
Remove some unnecessary header includes

GnomedDev:
- Removes unnecessary includes from the common files
- Removes duplicate includes from common and files that already included common
- Removes includes for `revolution.h` and `Common.h`(`h`) when not necessary


GnomedDev:
If this is wanted, it should be automated, and the extern "C" should be fixed by including it in the C files.

PR #693, opened 2023-04-25
Rename configure.py -> build.py and make responsible for invoking ninja 

GnomedDev:
This prevents issues of the `build.ninja` file becoming out of sync with reality by making `build.ninja` entirely temporary using the builtin `tempfile` library, then passing the path to this temporary file directly to `ninja`. This responsiblity change also allows for more complex logic to be built around the ninja invokation, since we control it.

For manual inspection of the resulting temporary file, `--dry` has been added to `build.py` which will produce the `build.ninja` in the current directory, although this will not be used for builds.

One issue that could arise is that people continue to run `ninja` with versions of `build.ninja` from before this change, although that should pretty quickly break and we can just tell them to use `build.py`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 build
 build.ninja
+build.ninja.old
 __pycache__
 out
 .vs
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 build
 build.ninja
-build.ninja.old
 __pycache__
 out
 .vs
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 build
 build.ninja
+build.ninja.old
```
stblr, line 3: No longer needed?

GnomedDev: Ah yes

PR #701, opened 2023-04-27
Cherry-pick / un-GPT ShaUtil

riidefi:


GnomedDev:
I dislike the Write/Read names, maybe make a class with a constructor and a toString method.

payload/sp/ShaUtil.hh
```diff
@@ -0,0 +1,16 @@
...
+
+#include <expected>
+#include <span>
+#include <string_view>
+
+#include "sp/FixedString.hh"
+
+namespace SP {
+
+FixedString<0x28 + 1> WriteSHA1(std::span<const u8, 0x14> bytes);
```
GnomedDev, line 13: Using FixedString is redundant as the length is always equal to the capacity. Just use a regular array.

payload/sp/ShaUtil.cc
```diff
@@ -0,0 +1,32 @@
...
+
+#include "HexParser.hh"
+
+namespace SP {
+
+FixedString<0x28 + 1> WriteSHA1(std::span<const u8, 0x14> bytes) {
+    FixedString<0x28 + 1> hex;
+
+    for (size_t i = 0; i < bytes.size(); ++i) {
+        snprintf(&hex.m_buf[i * 2], 3, "%02x", bytes[i]);
```
GnomedDev, line 14: Should this be split out into HexParser?

GnomedDev (changes requested):

riidefi:
I'll revert back to SHA1ToHex, SHA1FromHex

payload/sp/ShaUtil.cc
riidefi: I think we can drop HexParser and replace with https://en.cppreference.com/w/cpp/utility/to_chars

payload/sp/ShaUtil.hh
riidefi: That's true. But 4 bytes is probably not that much overhead; probably worth the trade-off for the cleaner interface? If we really wanted to be efficient we would be working with a templated output iterator anyway.

GnomedDev: What cleaner interface? The problem with splitting this out into it's own commit is it is void of context. In track-packs, I cannot remember a time where I needed to do anything but `.data()` on the array returned back.

riidefi:
I think neither of us knew of such a set of function's existence.

GnomedDev:
Oh, huh, yeah never knew that existed

riidefi:
Closing in favor of `hydro_bin2hex` and friends

PR #706, opened 2023-04-27
Workaround dol hashing issues on latest devkitpro

GnomedDev:
Seems like latest gcc breaks sha1 checking for little endian for some reason, leading to dol verification issues, this PR just works around it by hard coding big endian behavior but this should be looked into more.


stblr:
@GnomedDev did https://github.com/stblr/mkw-sp/commit/97502a8c1b01ee8287f590600c447115d838e91b fix it for you?

GnomedDev:
Yep

PR #707, opened 2023-04-28
Convert int and unsigned int to s32 and u32

GnomedDev:


vabold:
I hear basketball is a more fulfilling hobby than opening another 40+ file correctness PR.

Zachruff:
Stop with the attitude.

Gaberboo:
I agree with the conversions in file format structures, meh about small iterators, and I don't like it in function prototypes, particularly libc functions.

stblr (changes requested):
Library types should not be changed.

GnomedDev:
What do you mean by this? This isn't changing any types (just renaming them) and I'm not sure what your definition of library is here.

Gaberboo:
int and s32(int32_t) are different types that are equivelent in 32-bit PowerPC.

GnomedDev:
Okay? And I don't see any possible way we are compiling to something other than 32bit PPC for a very long time, if ever. Do you want me to add an assert of some kind to make sure the sizes are the same?

Gaberboo:
It's unlikely we will compile for anything else, even less likely that the target doesn't have 32-bit int, rather, it is rarely beneficial to intentionally make typings inconsistent with what the library originally used.

stblr:
It would be weird to have different types between declarations and definitions. For standard library functions, we only control the definition.

GnomedDev:
Isn't that most of the project's headers? Anyway, what are you defining as library definitions?

stblr:
An example would be `string.h`.

PR #717, opened 2023-05-06
Generate a symbol map file for all builds

MikeIsAStar:


generate_symbol_map.py
```diff
@@ -0,0 +1,60 @@
...
+if symtab_section is None:
+    sys.exit(f'There exists no \'.symtab\' section in the .elf file \'{args.elf_file_path}\'!')
+
+bad_symbols_regex = [
+    r'\.(.*)$',
+    r'^_GLOBAL__sub_[A-Z]_',
+    r'^patch_',
+    r'^replaced_',
+    r'^replacement_',
+]
```
GnomedDev, line 29: Can you wrap these strings in `re.compile` and then do `bad_symbol_regex.sub` below, that would pre-compile the regex parsers instead of compiling them each per symbol loop.

MikeIsAStar: Resolved in commit [987708c](https://github.com/stblr/mkw-sp/pull/717/commits/987708c486b203447d37d42c384eb8f21008dda9).

generate_symbol_map.py
```diff
@@ -0,0 +1,60 @@
...
+
+try:
+    smap_file_stream = open(args.smap_file_path, 'w')
+except:
+    sys.exit(f'Failed to open the symbol map file \'{args.smap_file_path}\'!')
+try:
+    for item in sorted(dictionary.items()):
+        smap_file_stream.write(f"0x{item[0]:X} {item[1]}\n")
+except:
+    sys.exit(f'Failed to write to the symbol map file \'{args.smap_file_path}\'!')
```
GnomedDev, line 57: Using bare except sounds kind of meh, especially when not using `with` for file handling, could this be refactored into a function and errors handled more specifically?

MikeIsAStar: > Using bare except sounds kind of meh, especially when not using `with` for file handling

File objects are flushed and closed once a script finishes executing.

> could this be refactored into a function and errors handled more specifically?

Which exceptions should be treated differently ?

GnomedDev: File objects being flushed and closed when a script closes is probably implementation defined behavior, and I've had issues before with Windows breaking due to this issue. It's mostly just good practice to use `with` and avoid bare `except`s, but specifically `KeyboardException` and maybe other internal stuff.

MikeIsAStar: Resolved in commit [088a389](https://github.com/stblr/mkw-sp/pull/717/commits/088a38970d0048d12046d18841b79e2ca45f7432).

build.py
```diff
@@ -51,10 +51,18 @@
...
 n.variable('wuj5', os.path.join('vendor', 'wuj5', 'wuj5.py'))
 n.newline()
 
+n.rule(
+    'generate_symbol_map',
+    command = f'{sys.executable} $generate_symbol_map $in $out',
+    description = 'SMAP $out',
+)
+n.newline()
+
```
stblr, line 65: I would maybe put it with the other code-related tools, after `ld`.

MikeIsAStar: Resolved in commit [eed02fd](https://github.com/stblr/mkw-sp/pull/717/commits/eed02fd6153febbcd5a7fb6ffbbdc88d1357ccc7).

stblr (approved):

PR #719, opened 2023-05-07
Increase the size of the symbol name buffer

MikeIsAStar:


stblr:
What's the reason for this change?

MikeIsAStar:
> What's the reason for this change?

This change allows for symbol names to be printed in full to the output port, even if they can not fit onto the television screen.

stblr (approved):

PR #729, opened 2023-05-11
Reimplement KartSaveState cleaner

### LINKED ISSUE #710: Drop save states
The implementation is a large hack that is only going to cause problems going forward. The plan is to hide it in 0.1.10 (#709) and remove it from the code base in 0.1.11.

GnomedDev:
This reimplements `KartSaveState` using a cleaner design, storing the minimum amount of data instead of trying to store everything (which contains pointers).

I've tested this doesn't store any pointers by moving `SaveStateManager` to static memory, then simply not resetting the `KartSaveState` when reloading the scene, and it works without crashing.

I could remove the headers added by `KartSaveState`, but they could be useful for future development, so opinions are wanted.

Closes #710 as the implementation is no longer such a hack.

### DIFF
--- a/payload/game/kart/KartBoost.hh
+++ b/payload/game/kart/KartBoost.hh
@@ -8,8 +8,11 @@ class KartBoost {
     friend class KartObjectProxy;
     friend class KartRollback;
 
+public:
+    KartBoost();
+    virtual ~KartBoost();
+
 private:
-    u8 _00[0x04 - 0x00];
     s16 m_timesBeforeEnd[6];
     u16 m_types;
     u8 _12[0x24 - 0x12];
--- a/payload/game/kart/KartBoost.hh
+++ b/payload/game/kart/KartBoost.hh
@@ -7,16 +7,22 @@ namespace Kart {
 class KartBoost {
     friend class KartObjectProxy;
     friend class KartRollback;
+    friend class KartSaveState;
 
 public:
     KartBoost();
-    virtual ~KartBoost();
+    virtual ~KartBoost() = delete;
+    virtual void dt(s32 type);
 
 private:
     s16 m_timesBeforeEnd[6];
     u16 m_types;
-    u8 _12[0x24 - 0x12];
+    f32 m_boostMultipler;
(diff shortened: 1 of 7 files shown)


payload/game/kart/KartBoost.hh
```diff
@@ -8,8 +8,11 @@ class KartBoost {
     friend class KartObjectProxy;
     friend class KartRollback;
 
+public:
+    KartBoost();
+    virtual ~KartBoost();
```
stblr, line 13: So this relies on compiler optimizations (devirtualization) for the destructor to be called?

I would suggest just copying the fields.

symbols.txt
```diff
@@ -651,6 +651,8 @@
 0x80581824 _ZN4Kart8KartMove9calcBlinkEv
 0x80584d58 KartMove_calcCannon
 0x805858ac KartMove_startKiller
+0x80588d28 _ZN4Kart9KartBoostC1Ev
+0x8057811c _ZN4Kart9KartBoostD1Ev
```
stblr, line 655: Ordering

payload/game/kart/KartBoost.hh
GnomedDev: How come?

stblr: GCC expects 2 destructors at 0x0 and 0x4, but the CW vtable only has one at 0x8.

stblr: It's possible to build a class that's compatible with both compilers. https://github.com/stblr/mkw-sp/blob/1ffe4da76271008adb5621252717745f5caa37e5/payload/game/ui/ctrl/CtrlRaceNameBalloon.cc#L164..L174

GnomedDev: Is it possible to retrofit an existing class though, without control over the constructor?

stblr: Yes, by rewriting the constructor.

stblr (approved):

Issue #667, opened 2023-04-03
"Course Select" and "Vehicle Restrict" options dont do anything on non-default settings.


Zachruff:

stblr:
Closed in #730 and #731.

PR #731, opened 2023-05-12
Implement vanilla vehicle restrictions

GnomedDev:
Part of the fix for #667, undoes the regression made by adding the new settings menu, but does not implement further filtering due to added complexity.

### DIFF
--- a/payload/game/ui/MultiTopPage.cc
+++ b/payload/game/ui/MultiTopPage.cc
@@ -63,9 +63,6 @@ void MultiTopPage::onInit() {
 void MultiTopPage::onActivate() {
     m_replacement = PageId::None;
 
-    auto *context = SectionManager::Instance()->globalContext();
-    context->m_vehicleRestriction = GlobalContext::VehicleRestriction::All;
-
     if (m_reset) {
         m_vsButton.selectDefault(0);
         m_instructionText.setMessage(3052);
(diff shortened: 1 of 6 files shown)


payload/sp/settings/ClientSettings.hh
```diff
@@ -127,10 +127,6 @@ enum class Vehicles {
     All,
     Karts,
     Bikes,
-    InsideDrift,
-    OutsideDrift,
-    Optimal,
-    Random,
```
stblr, line 133: Please comment them out like the rest.

payload/game/ui/SingleTopPage.cc
```diff
@@ -94,9 +118,6 @@ void SingleTopPage::onInit() {
 void SingleTopPage::onActivate() {
     m_replacement = PageId::None;
 
-    auto *context = SectionManager::Instance()->globalContext();
-    context->_74 = 2;
-
```
stblr, line 99: What happens in time trials?

payload/game/ui/MultiTopPage.cc
```diff
@@ -64,7 +64,7 @@ void MultiTopPage::onActivate() {
     m_replacement = PageId::None;
 
     auto *context = SectionManager::Instance()->globalContext();
-    context->_74 = 2;
+    context->m_vehicleRestriction = GlobalContext::VehicleRestriction::All;
```
stblr, line 67: Can you also implement the restrictions in multiplayer?

stblr (changes requested):

payload/game/ui/SingleTopPage.cc
GnomedDev: It only affects CPU drivers, so nothing happens.

stblr (approved):

PR #736, opened 2023-05-15
Rewrite the file 'Heap.c' in C++

MikeIsAStar:


payload/sp/Payload.cc
```diff
@@ -82,8 +84,8 @@ static void Init() {
 
     OSSetMEM1ArenaLo(Payload_getEnd());
 
-    Heap_RandomizeMEM1Heaps();
-    Heap_RandomizeMEM2Heaps();
+    SP::Heap::RandomizeMEM1Heaps();
+    SP::Heap::RandomizeMEM2Heaps();
```
vabold, line 88: As evidenced by `GlobalSettings::Init()` in the same function, the `SP` namespace in these calls is not required.

`SP::LogFile::Init()` can also be changed to `LogFile::Init()` to reflect this.

vabold (changes requested):

payload/sp/Payload.cc
MikeIsAStar: Resolved in commit [0c488ff](https://github.com/stblr/mkw-sp/pull/736/commits/0c488ffa1c175d438af30cac3588ce406b4915c8).

stblr (approved):

PR #738, opened 2023-05-15
Savestates: Sync more bullet state

GnomedDev:
Syncs buillets by just calling the activate and cancel methods on KartBullet. From quick testing, this works for `bullet -> not` and `not -> bullet` but goes in the wrong direction for a short time before recalculating for `bullet -> bullet`.


stblr (approved):
LGTM if conflicts are resolved.

stblr (approved):

PR #741, opened 2023-05-15
[RMCK] Correct the hash of the file 'main.dol'

MikeIsAStar:


stblr (approved):
I can confirm that this fixes it (on Dolphin).

PR #762, opened 2023-05-23
Fix insecure panic calls

riidefi:


MikeIsAStar:
Does there exist a scenario where the `panic` function is called with a format string that comes from an untrusted source ?

riidefi:
While I haven't been able to identify a specific instance, the game is making hundreds of calls that I've yet to thoroughly examine. And, that notwithstanding, calls like `panic("%s", "%s")` ought to be valid, but currently, aren't.

stblr (approved):

PR #763, opened 2023-05-23
Decompile ResourceManager::loadCourse

GnomedDev:
This is split off from #655, it loosly decompiles loadCourse while avoiding an extra buffer (writing directly to jobContext->filename instead of maintaining a buffer then copying) and asserts the file has loaded correctly, which prevents worse crashes if the filePath is wrong or something else is going wrong in course loading.

This PR also removes headers and symbols for `preloadCourseAsync` as it doesn't actually do anything with CourseCache stubbed.


payload/game/system/ResourceManager.cc
```diff
@@ -126,6 +123,47 @@ void ResourceManager::LoadGlobeTask(void *arg) {
     s_instance->loadGlobe(reinterpret_cast<u8 **>(arg));
 }
 
+MultiDvdArchive *ResourceManager::loadCourse(u32 courseId, EGG::Heap *heap, bool splitScreen) {
+    MultiDvdArchive *archive = m_archives[1];
```
riidefi, line 127: If we don't have an enum for these IDs, we can pull from the decomp here <https://github.com/riidefi/mkw/blob/master/source/game/system/MultiDvdArchive.hpp#L12>.

GnomedDev: Huh? Looking at the proper decomp for loadCourse doesn't even use that enum, what's it for?

riidefi: No worries. The fact that there are 9 resource channels and 9 entries in the array isn't a coincidence at all. The indices aren't meaningless numbers (in fact, changing any of them would break everything).
```cpp
enum ResourceChannelID {
  RES_CHAN_RACE_SYS, //!< [0] Localized: /Race/Common.szs and such
  RES_CHAN_COURSE,   //!< [1] Diff'd: Track file
  RES_CHAN_UI,       //!< [2] Localized: 2D UI
  RES_CHAN_FONT,     //!< [3] Fonts
  RES_CHAN_EARTH,    //!< [4] Globe (Async LZ)
  RES_CHAN_MII,      //!< [5] Mii Parts
  RES_CHAN_DRIVER,   //!< [6] Driver
  RES_CHAN_DEMO,     //!< [7] Cutscenes
  RES_CHAN_UI_MODEL, //!< [8] BackModel et al
  RES_CHAN_NUM,
};
```
Regarding the `loadCourse` in the decomp, that would be a skill issue on behalf of Mr. Bold if true.

riidefi: Apologies @GnomedDev, I didn't mean to come off that way. Let's focus on the issue at hand. The enum here helps in referring to specific resources without having to memorize or reference an index each time. It also improves code readability and maintenance. If there's any confusion about its application or how to use it, I'd be happy to clarify.

riidefi: That's perfect, nice work! We should probably have an equivalent to this, too?
```cpp
enum class JobContexts {
  Models,
  UI,
  Race,
  Mission,
  Competition,
  Kart, 
  System,
};
```

payload/game/system/ResourceManager.cc
```diff
@@ -126,6 +123,47 @@ void ResourceManager::LoadGlobeTask(void *arg) {
...
+
+    m_taskThread->request(&ResourceManager_doLoadTask, (void *)2, 0);
+    process();
+
+    u8 tries = 10;
+    while (!archive->isLoaded() && tries != 0) {
+        SP_LOG("Waiting for course archive to load... (%d tries left)", tries);
+        OSSleepMilliseconds(1000);
+        tries -= 1;
+    }
```
stblr, line 161: Isn't that redundant given that `process` is called?

payload/game/system/ResourceManager.hh
```diff
@@ -4,9 +4,13 @@
 #include "game/system/MultiDvdArchive.hh"
 
 #include <egg/core/eggExpHeap.hh>
+#include <egg/core/eggTaskThread.hh>
 
 namespace System {
 
+// Loads the archive setup in ResourceManager->m_jobContexts
+extern void ResourceManager_doLoadTask(void *contextIdx);
```
stblr, line 12: This should be a private static function. You can look at `SaveManager` for examples.

payload/game/system/ResourceManager.cc
GnomedDev: I'm really not sure how this works, to be honest, but I know that process can return without loading the archive so added this to make sure it isn't just taking a while.

stblr: When `process` returns, it means that all tasks have finished running. Thus only the assert should be needed.

stblr (approved):

PR #779, opened 2023-05-24
 Add controls to `/ui_info` command

riidefi:


riidefi:
Repurposed as "Add controls to /ui_info command"

payload/game/ui/ControlGroup.cc
```diff
@@ -2,10 +2,29 @@
...
+#include <cstring>
+
 namespace UI {
 
 ControlGroup::~ControlGroup() {
     dt(-1);
 }
 
+void ControlGroup::logDebug(int depth) {
+    const int MAX_DEPTH = 16;
```
stblr, line 14: Could you add a max depth check?

riidefi: Done

stblr: Should be `>=` I think.

riidefi: If depth==MAX_DEPTH, wouldn't everything work as expected?

stblr: You are right, that `memset` confused me.

stblr (approved):

PR #780, opened 2023-05-26
Stub all ParameterFile functions

GnomedDev:
This allows for safe usage of the first 0x16 bytes of RaceConfig (after the vtable).


payload/game/system/RaceConfig.hh
```diff
@@ -127,7 +127,9 @@ public:
 private:
     REPLACE static void ConfigurePlayers(Scenario &scenario, u32 screenCount);
 
+    // Was used for ParameterFile, but now unused and free for replacement.
     u8 _0004[0x0020 - 0x0004];
```
vabold, line 131: The destructor must be replaced before it is safe to use 0x04 - 0x1c. `DestroyInstance` must be replaced before it is safe to use 0x1c - 0x20.

vabold (changes requested):

payload/game/system/RaceConfig.hh
stblr: There is also a vtable access in `RaceConfig::init`.

vabold: The vtable access in `init` is fine because the dereference at 0x0 is valid. We simply stub the function.

GnomedDev:
This doesn't seem worth it.

PR #786, opened 2023-05-29
Page: Use `Sound::SoundId` for `setAnimSfxIds` parameters

riidefi:


payload/game/ui/Page.hh
```diff
@@ -8,6 +8,10 @@
 #include "game/ui/TypeInfo.hh"
 #include "game/ui/UIControl.hh"
 
+namespace Sound {
+enum class SoundId;
+} // namespace Sound
+
```
stblr, line 14: I don't think that's needed.

riidefi: As in, please include `SoundId` in this header?

stblr: It's already included in `Page.hh`.

riidefi: You're right

stblr (approved):

PR #788, opened 2023-05-29
Page ID extension

riidefi:


payload/game/ui/Page.cc
```diff
@@ -64,4 +64,9 @@ void Page::logControlsDebug() {
     m_controlGroup.logDebug(1);
 }
 
+void Page::deactivate() {
+    onDeactivate();
+    m_state = State::State1;
+}
```
vabold, line 70: Seems unnecessary.

payload/game/host_system/SystemManager.hh
```diff
@@ -32,7 +32,11 @@ private:
     u8 _0000[0x0058 - 0x0000];
     u32 m_aspectRatio;
     u8 _005c[0x0070 - 0x005c];
+
+public:
     u32 m_launchType;
+
+private:
```
vabold, line 39: Maybe a getter function would be preferred here.

payload/game/ui/Section.cc
```diff
@@ -496,6 +496,57 @@ void Section::logDebuggingInfo(bool verbose) {
...
+extern "C" void Page_unfocusMaybe(Page *);
+
+void Section::deinit() {
+    // Since we zero out the arrays, we can trust REPLACED(deinit) does not double-free.
+    for (u32 i = m_activePageCount; i > 0; --i) {
+        auto &page = m_activePages[i - 1];
+        Page_unfocusMaybe(page);
+        page->deactivate();
+        page = nullptr;
+    }
```
vabold, line 517:
```suggestion
    cullActivePages(0);
```
We can use the existing base game helper function at 806232e8.

The behavior here is also incorrect, as we do not set the page to a nullptr in this loop.

payload/game/ui/Section.cc
```diff
@@ -496,6 +496,57 @@ void Section::logDebuggingInfo(bool verbose) {
...
+    for (auto &page : m_systemPages) {
+        if (page == nullptr) {
+            continue;
+        }
+        Page_unfocusMaybe(page);
+        page->onDeinit();
+        // delete page will call GCC dtor which will be null
+        page->dt(1);
+        page = nullptr;
+    }
```
vabold, line 546: This loop is not required.

payload/game/ui/Section.cc
```diff
@@ -496,6 +496,57 @@ void Section::logDebuggingInfo(bool verbose) {
...
+    // Since we zero out the arrays, we can trust REPLACED(deinit) does not double-free.
+    for (u32 i = m_activePageCount; i > 0; --i) {
+        auto &page = m_activePages[i - 1];
+        Page_unfocusMaybe(page);
+        page->deactivate();
+        page = nullptr;
+    }
+    m_activePageCount = 0;
+    for (auto &page : m_pages) {
+        if (page == nullptr) {
```
vabold, line 520:
```suggestion
    for (auto *&page : m_pages) {
        if (!page) {
```
Along with similar changes in this function.

I think an implicit nullptr check is fine in this context as there is no confusion - the page either exists or it doesn't.

payload/game/ui/SectionManager.cc
```diff
@@ -29,9 +30,24 @@ GlobalContext *SectionManager::globalContext() {
     return m_globalContext;
 }
 
-void SectionManager::createSection() {
-    REPLACED(createSection)();
+extern "C" void RegisteredPadManager_8061B5A4(void *);
+extern "C" void RegisteredPadManage_onCreateSection(void *, bool);
```
vabold, line 34:
```suggestion
extern "C" void RegisteredPadManager_onCreateSection(void *, bool);
```

payload/game/ui/PageId.hh
```diff
@@ -209,6 +209,24 @@ enum class PageId {
...
+    EXT_MAXEXCLUSIVE__,
 };
 
+static_assert(static_cast<size_t>(PageId::EXT_MAXEXCLUSIVE__) >=
+        static_cast<size_t>(PageId::EXT_MININCLUSIVE__));
+
+constexpr size_t ExtensionPageCount() {
+    return static_cast<size_t>(PageId::EXT_MAXEXCLUSIVE__) -
+            static_cast<size_t>(PageId::EXT_MININCLUSIVE__);
+}
```
vabold, line 228: Can we put this in a `UI::PageId` namespace or something similar?

payload/game/ui/SectionManager.cc
```diff
@@ -29,9 +30,24 @@ GlobalContext *SectionManager::globalContext() {
...
+    }
+    auto nextId = m_nextSectionId;
+    m_currentAnimDir = m_nextAnimDir;
+    m_nextSectionId = UI::SectionId::None;
+    m_changeTimer = 0;
+    m_nextAnimDir = 0;
+    m_transitionFrame = -1;
+    m_state = 0;
+    bool race = Section::GetSceneId(nextId) == static_cast<u32>(System::RKSceneID::Race);
+    RegisteredPadManage_onCreateSection(&m_registeredPadManager, !race);
```
vabold, line 48:
```suggestion
    RegisteredPadManager_onCreateSection(&m_registeredPadManager, !race);
```

vabold (changes requested):

vabold:
Nice work with the SectionManager field finding and the extension. Does Clang handle the `offsetof` correctly without pragmas?

payload/game/ui/Section.cc
riidefi: I think I prefer `nullptr`

payload/game/ui/PageId.hh
riidefi: This seems tacky

vabold: This is not an uncommon practice in the repository. The `SP` namespace notably has `Channel`, `IOSDolphin`, `MapFile`, etc. with similar functionality.

payload/game/ui/Section.cc
riidefi: I don't think the behavior is incorrect in this context. We can use the helper, though.

vabold: If the page is a nullptr, `onDeinit` is never called in the next loop.

payload/game/ui/PageId.hh
riidefi: While it is true that the repository has many namespaces, I am not sure how I feel about shadowing a typename with a namespace

payload/game/ui/Section.cc
riidefi: This is irrelevant because `m_activePageCount` will be zero in all cases here, so the values of `m_activePages` will never be inspected. With respect to the other arrays, we rely on the zeroing to avoid a double-free bug

vabold: Oh, sorry, I misunderstood that setting the page to a nullptr would affect `m_pages` and not `m_activePages`. Yes, you're right.

payload/game/ui/Section.cc
vabold: Should we at least have `auto *&` for transparency about the type?

payload/game/ui/PageId.hh
vabold: Yeah, you're right. Perhaps it is best scoped in `Page` as a static function. I'm hesitant to leave such a specific function in such a generic namespace.

riidefi: I think I like this less, since we should keep the enum functions and static checks as close to the enum as possible for maintainability. Another option could be to add an enum member `PageId::EXT_Count = ...`, but I can't really see much advantage to it.

payload/game/ui/Section.cc
riidefi: Yes

riidefi:
I haven't checked with Clang, but if not, we can just conditionally disable that check when we get to it.

payload/game/ui/Section.hh
```diff
@@ -85,18 +85,35 @@ private:
...
     nw4r::lyt::DrawInfo m_drawInfo;
     u8 _3e4[0x3f8 - 0x3e4];
     Vec2<f32> m_scaleFor;
     u8 _400[0x408 - 0x400];
+
+public: // for static assert
+    std::array<Page *, ExtensionPageCount()> m_pagesEXT{};
 };
-static_assert(sizeof(Section) == 0x408);
+// sizeof(Section) == 0x408 + sizeof(Section::m_pagesEXT) breaks when ExtensionPageCount() is 0
```
stblr, line 112: What about `sizeof(Section == 0x408 + sizeof(Page *) * ExtensionPageCount()`?

riidefi: Since the array is not declared `[[no_unique_address]]`, it must be at least 1 byte long when it is *instantiated* in the class, even though it is 0 bytes long itself. The correct expression would be `sizeof(Section) == 0x408 + std::max(sizeof(Section::m_pagesEXT), 1)` which is not great

payload/game/ui/PageId.hh
vabold: > I am not sure how I feel about shadowing a typename with a namespace

Good intuition - it doesn't compile.

Keeping it close to the enum and having a stricter scope does not seem feasible. I'm inclined to prioritize the latter goal. If we cannot keep it close to the enum, we should instead keep it close to its usecase in `Section`.

riidefi: > Another option could be to add an enum member PageId::EXT_Count = ..., but I can't really see much advantage to it.
This would work and maximize both of your goals, but I still am not convinced there is any deficiency with the current approach.

vabold: The deficiency with the current approach is that the function's scope does not match the function's versatility.

Would that member need to be updated manually when a new page is introduced?

riidefi: No, such an approach would not need to be manually updated; an enum value can be any constant expression, even the result of a `constexpr` function.

It's fairly common to immediately place constexpr helpers immediately after an enum in the same scope, e.g. [here](https://github.com/llvm/llvm-project/blob/b936816fb305cc16d2d1cd6d424c08c39a681a32/clang/include/clang/Basic/Specifiers.h#L203) or [here](https://github.com/google/draco/blob/9f856abaafb4b39f1f013763ff061522e0261c6f/src/draco/compression/mesh/mesh_edgebreaker_shared.h#L53).

vabold: Yes, but the enum class is a different scope compared to the enums in these examples. The equivalent of your examples would be `enum PageId` which is accessed as `UI::PAGE_EMPTY` or something similar.

riidefi: No, it's written `Enum_Value` and not `Enum::Value` because `enum class` is a modern C++ feature. In principle, the practice of putting helper methods right below the enum, in the same namespace, is standard practice. If it's good enough for LLVM, it should be good enough for SP.

vabold: @stblr What are your thoughts?

riidefi: Enum class [here](https://github.com/llvm/llvm-project/blob/af8e3861025f3c931cc67ced86d6bd8c939e6fc0/clang/lib/CodeGen/CGValue.h#L130) and [here](https://myanimelist.net/anime/4280/Kara_no_Kyoukai_Movie_4__Garan_no_Dou).

vabold: I'm not deeply familiar with the C++ standard but I do not think "Kara no Kyoukai Movie 4: Garan no Dou" is an enum class.

riidefi:
```cpp
/// The source of the alignment of an l-value; an expression of
/// confidence in the alignment actually matching the estimate.
enum class AlignmentSource {
  /// The l-value was an access to a declared entity or something
  /// equivalently strong, like the address of an array allocated by a
  /// language runtime.
  Decl,

  /// The l-value was considered opaque, so the alignment was
  /// determined from a type, but that type was an explicitly-aligned
  /// typedef.
  AttributedType,

  /// The l-value was considered opaque, so the alignment was
  /// determined from a type.
  Type
};

/// Given that the base address has the given alignment source, what's
/// our confidence in the alignment of the field?
static inline AlignmentSource getFieldAlignmentSource() {
  // For now, we don't distinguish fields of opaque pointers from
  // top-level declarations, but maybe we should.
  return AlignmentSource::Decl;
}
```
If it works for LLVM it works for SP. Marking as resolved.

payload/game/ui/Section.hh
riidefi: Given `sizeof(Section == 0x408 + sizeof(Page *) * ExtensionPageCount())` is not a viable alternative, marking as resolved.

riidefi: This should also be standardized in C++26 https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2883r0.pdf

payload/game/ui/SectionManager.cc
```diff
@@ -30,8 +31,20 @@ GlobalContext *SectionManager::globalContext() {
...
+    auto nextId = m_nextSectionId;
+    m_currentAnimDir = m_nextAnimDir;
+    m_nextSectionId = UI::SectionId::None;
+    m_changeTimer = 0;
+    m_nextAnimDir = 0;
+    m_transitionFrame = -1;
+    m_state = 0;
+    bool race = Section::GetSceneId(nextId) == static_cast<u32>(System::RKSceneID::Race);
+    m_registeredPadManager.onCreateSection(!race);
+    m_currentSection = new Section;
```
stblr, line 46: Looks like the vanilla ctor is not called here, is that intended?

riidefi: Are you sure? I would expect it to call our GCC default ctor. That is itself an issue, though.

stblr: Yeah that's what I mean.

riidefi: Fixed

vabold (approved):

stblr (approved):

PR #804, opened 2023-06-01
Ensure that battle stages do not appear on the leaderboards

MikeIsAStar:


vabold (approved):

stblr (changes requested):
This breaks thumbnail preloading.

The `race` part of the filter is already overridden if we are in the `Rankings` section, we should simply do the same for the `battle` part.

vabold:
> The `race` part of the filter is already overridden if we are in the `Rankings` section, we should simply do the same for the `battle` part.

Instead of incorporating multiple potential usecases over time, we instead solve the problem now by extending the `filter` helper function to allow filters as arguments.

MikeIsAStar:
> This breaks thumbnail preloading.
> 
> The `race` part of the filter is already overridden if we are in the `Rankings` section, we should simply do the same for the `battle` part.

Resolved by commit [1f5cc52](https://github.com/stblr/mkw-sp/pull/804/commits/1f5cc52658c8928ecb22bf5d5ac676f085043ac1).

stblr (approved):

PR #805, opened 2023-06-01
Pregenerate course selection

### LINKED ISSUE #761: Random Tracks Repeat Same Tracks
Unlike the vanilla game, it is possible to get repeat tracks in the same set of tracks for VS Mode when set to Random.

GnomedDev:
Loosely split from #655. Closes #761.


payload/game/system/RaceConfig.cc
```diff
@@ -118,6 +145,14 @@ void RaceConfig::applyCPUMode() {
     }
 }
 
+void RaceConfig::initRace() {
+    REPLACED(initRace());
+
+    if (m_menuScenario.gameMode != GameMode::Awards) {
+        m_raceScenario.courseId = *m_courseOrder[m_currentCourse];
+    }
```
stblr, line 153: How is the title screen demo handled?

payload/game/system/RaceConfig.cc
```diff
@@ -177,32 +212,66 @@ void RaceConfig::ConfigurePlayers(Scenario &scenario, u32 screenCount) {
...
+    auto courseCount = courseDatabase.count(filter);
+
+    auto raceCount = UI::SectionManager::Instance()->globalContext()->m_matchCount;
+    for (u8 i = 0; i < raceCount; i += 1) {
+        Registry::Course randCourse;
+
+        do {
+            auto courseIdx = hydro_random_uniform(courseCount);
+            randCourse = courseDatabase.entry(filter, courseIdx).courseId;
+        } while (m_courseOrder.contains(randCourse));
```
stblr, line 247: What happens if `raceCount` is larger than `courseCount`?

payload/game/system/GhostFile.cc
```diff
@@ -259,7 +259,7 @@ bool IsValid(const u8 *raw, u32 size) {
         return false;
     }
 
-    if (header->courseId >= 0x20) {
+    if (static_cast<u32>(header->courseId) >= 0x20) {
```
riidefi, line 262: Move to `CourseIdIsValid`

payload/game/ui/SingleTopPage.cc
```diff
@@ -86,6 +86,8 @@ void SingleTopPage::onInit() {
     m_taButton.selectDefault(0);
     m_instructionText.setMessage(3051);
 
+    System::RaceConfig::Instance()->clearCourses();
```
stblr, line 89: What about `MultiTopPage`?

payload/game/system/RaceConfig.cc
```diff
@@ -177,32 +212,66 @@ void RaceConfig::ConfigurePlayers(Scenario &scenario, u32 screenCount) {
...
+
+bool RaceConfig::generateRandomCourses() {
+    if (getCourseSelection() != SP::ClientSettings::CourseSelection::Random) {
         return false;
     }
 
-    if (setting != SP::ClientSettings::CourseSelection::Random) {
+    SP::CourseDatabase::Filter filter;
+    filter.battle = m_menuScenario.isBattle();
+    filter.race = !filter.battle;
```
riidefi, line 235: {.battle = ..., .race = ... };

payload/game/system/RaceConfig.cc
```diff
@@ -177,32 +212,66 @@ void RaceConfig::ConfigurePlayers(Scenario &scenario, u32 screenCount) {
...
+}
+
+bool RaceConfig::generateOrderedCourses(u16 startingIndex) {
+    if (getCourseSelection() != SP::ClientSettings::CourseSelection::InOrder) {
         return false;
     }
 
+    SP::CourseDatabase::Filter filter;
+    filter.battle = m_menuScenario.isBattle();
+    filter.race = !filter.battle;
```
riidefi, line 262: ditto

stblr (changes requested):

payload/game/system/GhostFile.cc
GnomedDev: uh, do you want me to create a new function for that static_cast or do some extra checks, because extra checks sounds out of scope.

riidefi: Do something similar to what hnery did, with a switch case CRS_0 ... CRS_32 essentially

GnomedDev: Out of scope for this PR then.

riidefi: Retyping `courseId` to an enum would be the out-of-scope change, if any. Notwithstanding, if you want to make that change in this PR, it should be done correctly.

GnomedDev: I know it is somewhat out of scope now I've finished it, but started off as a small change to improve type safety. I can split it off into a separate PR if you want, but I don't think we should be making a passable situation worse by adding entirely new validity checking logic.

riidefi: I would add this to Registry.cc
```cpp
bool IsControllerValid(Controller controller) {
    switch (controller) {
    case Controller::WiiWheel... Controller::GameCube:
        return true;
    default:
        return false;
    }
}
```
I would expect something like this in any PR that has retyping courseId as a goal

payload/game/system/RaceConfig.cc
GnomedDev: Good catch, seems like in testing it always does Luigi's circuit, I'll see how to fix.

payload/game/system/RaceConfig.cc
GnomedDev: Currently it would infinite loop here, how does the base game handle it?

GnomedDev: Looks like the base game doesn't handle this, for battle it generates 10 courses and for race it generates 32 courses.

payload/game/ui/CourseSelectPage.cc
```diff
@@ -132,6 +132,11 @@ void CourseSelectPage::onDeinit() {
 
 void CourseSelectPage::onActivate() {
     m_replacement = PageId::None;
+
+    auto section = SectionManager::Instance()->currentSection()->id();
+    if (section != SectionId::SingleSelectVSCourse && section != SectionId::SingleSelectBTCourse) {
+        System::RaceConfig::Instance()->clearCourses();
```
stblr, line 138: This won't be called if course selection is randomized.

GnomedDev: It will if it's ordered.

riidefi:
Would it make more sense to use the arrays in `GlobalContext`? Or if we want, extend GlobalContext with a new array?
| Offset | Name |
|--------|--------|
| GlobalContext+0x78   |  u32 mTrackOrder[32]; | 
| GlobalContext+0xF8   |  u32 mTrackOrderLen; | 
| GlobalContext+0xFC | u32 mBattleOrder[10]; |

| Address | Name |
|----------|--------|
| 808606A4 | GlobalContext_getCourseMatch(u32 index) |
| 808606D0 | GlobalContext_getBattleMatch(u32 index) |

I think the idea would be to keep the course order in `GlobalContext` and let `RaceConfig` stay a mostly dumb POD mostly concerned with the current race?

It seems a bit weird to have `RaceConfig` encroach on `GlobalContext`'s duties like this?

GnomedDev:
No, because of the reasons I said in the dev channel before you commented that.

riidefi:
LGTM! I didn't realize we still had SectionManager.c around to complicate things

riidefi (approved):

GnomedDev:
Yeah, I was able to very much stub it, but we still need that old-style patch for the GlobalContext allocation.

payload/game/ui/CourseSelectPage.cc
GnomedDev: Fixed this, after remembering about the discussion on discord.

GnomedDev:
Fixed merge conflicts.

PR #811, opened 2023-06-02
Add assertion to GroupContainer_FindGroupByName

riidefi:
This was useful in discovering a mislabled `pane` field is actually a GroupContainer name and should help make the UI system more friendly to work with.

### DIFF
--- a/payload/nw4r/lyt/lyt_GroupContainer.cc
+++ b/payload/nw4r/lyt/lyt_GroupContainer.cc
@@ -1,0 +1,10 @@
+#include "Common.hh"
+
+extern "C" void *REPLACED(GroupContainer_FindGroupByName)(void *a, const char *b);
+extern "C" REPLACE void *GroupContainer_FindGroupByName(void *a, const char *b) {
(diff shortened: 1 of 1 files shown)


payload/nw4r/lyt/lyt_GroupContainer.cc
```diff
@@ -0,0 +1,10 @@
+#include "Common.hh"
+
+extern "C" void *REPLACED(GroupContainer_FindGroupByName)(void *a, const char *b);
+extern "C" REPLACE void *GroupContainer_FindGroupByName(void *a, const char *b) {
+    void *r = REPLACED(GroupContainer_FindGroupByName)(a, b);
+    if (r == 0) {
```
stblr, line 6: `!r` or `r == nullptr`

stblr (changes requested):

payload/nw4r/lyt/lyt_GroupContainer.cc
riidefi: Fixed

stblr (approved):

PR #815, opened 2023-06-03
[Console] Export `TextWriter` API

riidefi:
This is useful in its own right


stblr:
Does that mean that we have 3 text rendering APIs in mkw-sp?

stblr:
Why is it still in `Console.h` if it's standalone?

riidefi:
> Does that mean that we have 3 text rendering APIs in mkw-sp?

It does appear that we have 3 text rendering APIs in mkw-sp. However, if someone reverse engineers the `nw4r::ut::TextWriter` API, considering its similarity in domain (being GX-based), we could consider phasing this one out--or, in the interim, reimplement this API using `ut::TextWriter`. Notwithstanding, making this API a visible dependency is a good step for the repository, especially if it is ultimately redundant.

> Why is it still in Console.h if it's standalone?

I understand where you're coming from. I agree, it can seem a bit odd to keep it in `Console.h`, even though it's technically standalone. The reason behind this is because it utilizes `UI::GlyphRenderer` from the game, so it does have a dependency there. I share your concern about creating circular dependencies between the SP and Game folders, which we want to avoid for better code organization.

Taking your point into account, moving it to its own file does sound like a sensible idea. By exporting it, we're already heading towards that goal.

riidefi:
I have extracted it to its own file `TextWriter.c`

stblr (approved):

PR #817, opened 2023-06-03
Storage.hh: Let `NodeInfo`, `DirHandle` `FileHandle` be value-comparable

riidefi:


payload/sp/storage/Storage.hh
```diff
@@ -78,6 +80,8 @@ public:
     DirHandle &operator=(DirHandle &&);
     ~DirHandle();
 
+    bool operator==(const DirHandle &) const = default;
+
```
stblr, line 84: What about `FileHandle`?

payload/sp/storage/Storage.hh
```diff
@@ -28,6 +28,8 @@ struct NodeInfo {
     u64 size;
     OSTime tick = 0;
     wchar_t name[255 + 1];
+
+    bool operator==(const NodeInfo &) const = default;
```
stblr, line 32: What's the default comparison for the `name` array?

stblr (changes requested):

payload/sp/storage/Storage.hh
riidefi: `memcmp`

riidefi: Changed to `!wcsncmp`

payload/sp/storage/Storage.hh
riidefi: You're right, added.

stblr (approved):

PR #819, opened 2023-06-03
Rewrite ObjPylon01.c in C++

riidefi:


riidefi:
Only done to allow `RaceConfig.h` to be deleted. A follow-up PR could include a proper rewrite

payload/game/obj/ObjPylon01.cc
```diff
@@ -0,0 +1,16 @@
+extern "C" {
+#include "ObjPylon01.h"
+}
+
+#include "../system/RaceConfig.hh"
```
vabold, line 5:
```suggestion
#include "game/system/RaceConfig.hh"
```

payload/game/obj/ObjPylon01.cc
```diff
@@ -0,0 +1,16 @@
...
+extern "C" u32 my_ObjPylon01_vf_b0(const ObjPylon01 *self) {
+    auto *cfg = System::RaceConfig::Instance();
+    if (cfg->raceScenario().players[0].type != System::RaceConfig::Player::Type::Ghost &&
+            self->playerId != 0) {
+        return 1; // Transparent
+    }
+
+    return 12; // Opaque
+}
+PATCH_B(ObjPylon01_vf_b0, my_ObjPylon01_vf_b0);
```
vabold, line 16: Please use the `REPLACE` macro.

vabold (changes requested):

payload/game/obj/ObjPylon01.cc
riidefi: It was this way in the original file.

payload/game/obj/ObjPylon01.cc
riidefi: It was this way in the original file.

riidefi:
Closed in favor of [#820](https://github.com/camzillajs101/garfield-cursed-image/blob/master/app/assets/images/amalgam.jpg?raw=true)

PR #820, opened 2023-06-03
Rewrite RaceConfig in C++

vabold:
Pending testing.
- [x] Dolphin
- [x] Console

Closes #819.


payload/game/obj/ObjPylon01.cc
```diff
@@ -0,0 +1,12 @@
+#include "ObjPylon01.hh"
+
+#include "game/system/RaceConfig.hh"
+
+namespace Geo {
+
+u32 ObjPylon01::vf_b0() {
+    const auto &player = System::RaceConfig::Instance()->raceScenario().players[0];
+    return player.type != System::RaceConfig::Player::Type::Ghost && m_playerId != 0 ? 1 : 12;
```
riidefi, line 9: Please bring back the "Transparent" and "Opaque" comments. It's unclear what this patch even does now.

payload/game/obj/ObjPylon01.hh
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "game/obj/Obj.hh"
+
+namespace Geo {
+
+class ObjPylon01 : public Obj {
+public:
```
riidefi, line 8:
```diff
+ ObjPylon01() = delete;
REPLACE u32 vf_b0();
+ // ...
```
to make clear that the vtable is incomplete and the default ctor is invalid.

GnomedDev: Could just define as `ObjPylon01();` to error at link time if not found, that is what I do when I want to convey that.

riidefi:
Tested 2 ghosts on Dolphin; this patch works. Behavior should not be different on Console.

payload/game/obj/ObjPylon01.hh
riidefi: This also works.

vabold: Resolved in cd88dea.

payload/game/obj/ObjPylon01.cc
vabold: Resolved in cd88dea.

payload/game/obj/ObjPylon01.cc
```diff
@@ -0,0 +1,15 @@
+#include "ObjPylon01.hh"
+
+#include "game/system/RaceConfig.hh"
+
+namespace Geo {
+
+Obj::PixelMode ObjPylon01::vf_b0() {
+    using Mode = PixelMode;
+    using Type = System::RaceConfig::Player::Type;
+
```
vabold, line 10: This is the result of formatter wrangling. I was originally going to use `constexpr u32 TRANSPARENT = 1;` and etc, but riidefi recommended defining the enum class. Because the enum class splits the ternary into two separate lines, I opt to add `using Mode = PixelMode;` to achieve both good formatting and optimized instructions.

stblr: An `if` block would be easier to read.

vabold: Resolved in 265a788.

payload/game/obj/Obj.hh
```diff
@@ -16,6 +16,11 @@ public:
     virtual void init() = 0;
     // ...
 
+    enum class PixelMode {
+        Transparent = 1,
```
riidefi, line 20: nit: Should be `Translucent` (XLU), in NW4R `Transparent` refers to stencil-clip

vabold: No thanks.

GnomedDev (approved):
LGTM

riidefi (approved):

stblr (approved):

PR #821, opened 2023-06-04
Move updateserver::Stream to netprotocol

GnomedDev:
This moves the N key negotiation to `netprotocol` and makes `netprotocol::AsyncStream` agnostic over negotiation methods.

`updateserver` was also slightly cleaned up in the process, following some clippy comments, but nothing that should change behaviour.

Needs testing:
- [ ] updateserver
- [x] matchmaking
- [x] gameserver

In future, this allows for moving matchmaking to use N key negotiation, as it currently uses XX.


GnomedDev:
Where is this documented?

PR #822, opened 2023-06-04
Rewrite RaceManager in C++

riidefi:
Tested on Dolphin.

Closes #819.


riidefi:
Takes a different approach by reimplementing `RaceConfigScenario_resetGhostPlayerTypes` as a true C++ member function, replacing the `RaceConfig` constructor directly rather than a static initializer. Care has been taken to explicitly delete default, copy and move constructors when their use would cause memory safety vulnerabilities, e.g. dangling pointers.

PR #827, opened 2023-06-06
[Storage] Add `std::directory_iterator` analog

riidefi:


GnomedDev:
Just wondering, if this is an analog to a stdlib thing, could we get the stdlib fs code working on Wii or is it not that portable?

riidefi:
That's a good question. I'm not sure how practical it'd be, here's the GCC one, for instance:
```cpp
class recursive_directory_iterator
{
  // ...
  recursive_directory_iterator(const path& __p)
    : recursive_directory_iterator(__p, directory_options::none, nullptr)
  { }
  // ...
  directory_options  options() const noexcept;
  int                depth() const noexcept;
  bool               recursion_pending() const noexcept;
  // ...
  void pop(error_code&);
  void disable_recursion_pending() noexcept;
  // ...
  struct _Dir_stack;
  std::__shared_ptr<_Dir_stack> _M_dirs;
  // ...
  recursive_directory_iterator& __erase(error_code* = nullptr);
};
```
The standard filesystem library does a bunch of allocations (not just the ARC pointer here, but also all paths are just `std::string`s), which might not be the most appropriate for the game?
That being said, we could certainly implement the Posix APIs GCC is going to call, and presumably our code would just work.

riidefi:
Addresses half of #842

PR #832, opened 2023-06-13
Implement utils for working with Sha

GnomedDev:
Split off from #655

### DIFF
--- a/payload/sp/ShaUtil.cc
+++ b/payload/sp/ShaUtil.cc
@@ -1,0 +1,19 @@
+#include "ShaUtil.hh"
+
+#include <vendor/libhydrogen/hydrogen.h>
+
+std::array<char, 0x28 + 1> sha1ToHex(const Sha1 &sha1) {
+    std::array<char, 40 + 1> out;
+    hydro_bin2hex(out.data(), out.size(), sha1.data(), sha1.size());
+    return out;
+};
+
+std::expected<Sha1, const char *> sha1FromHex(std::string_view sv) {
+    Sha1 out;
+    u32 val = hydro_hex2bin(out.data(), out.size(), sv.data(), sv.size(), nullptr, nullptr);
+    if (val == out.size()) {
+        return out;
+    } else {
+        return std::unexpected("Failed to parse sha1 from hex!");
+    }
+}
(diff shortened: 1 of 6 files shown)


riidefi:
I think this is fine, but the annoying thing with C++ is that a std::array reference isn't like a Rust arrayref, so this code sort of requires that everyone is locked into the Sha1 struct. Perhaps we could provide two aliases? i.e.
```cpp
using Sha1 = std::array<u8, 0x14>;
using Sha1Ref = std::span<u8, 0x14>;
```

GnomedDev:
Being locked into the `Sha1` struct is mostly the point, for type safety.

riidefi:
Okay--while I personally think this is quite brittle and odd for a function on a read-only span of 20 bytes requiring the programmer to memcpy them into a specific structure (when reading them a file is a use case), I can agree that the security footguns of the current approach (with a raw pointer) are worse. I've approved this change.

riidefi (approved):

payload/sp/ShaUtil.hh
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <Common.hh>
+
+#include <expected>
+#include <string_view>
```
vabold, line 6:
```suggestion
#include <expected>
#include <string_view>

#include <Common.hh>
```

payload/sp/ShaUtil.cc
```diff
@@ -0,0 +1,18 @@
+#include "ShaUtil.hh"
+#include <vendor/libhydrogen/hydrogen.h>
```
vabold, line 2: While you're fixing includes:
```suggestion
#include "ShaUtil.hh"

#include <vendor/libhydrogen/hydrogen.h>
```

vabold (changes requested):

payload/sp/ShaUtil.hh
GnomedDev: Huh, that's weird, I could swear it was common before stdlib, but I've gone back and checked and nope. There are probably much more examples in the project of incorrect formatting that a tidy check should really be written for it.

GnomedDev:
Solved conflicts and header ordering issues.

vabold (approved):

Issue #829, opened 2023-06-06
DISCUSSION: Upgrade to clang-format 16


vabold:
In clang-format 16, `AlignTrailingComments` was changed from being True/False into a [more configurable option](<https://clang.llvm.org/docs/ClangFormatStyleOptions.html#aligntrailingcomments>). This introduces problems for users of recent clang-format versions, as there are two paths forward.

1) Locally change `AlignTrailingComments` and ensure `.clang-format` changes are not committed
2) Downgrade the local clang-format version to ensure compatibility

Neither option is preferable for users, and given most contributors I've talked to seem to use clang-format 16 and use option 1 as their go-to strategy, I feel that deprecating clang-format 15 is worth considering.

stblr:
At the moment, clang-format 16 is not available in Ubuntu 22.04.

PR #836, opened 2023-06-15
Remove AlignTrailingComments format option

GnomedDev:
This option existing in the config seems to cause clang-format 16 to error with
```
.clang-format:9:24: error: not a mapping
AlignTrailingComments: True
                       ^~~~
Error reading .clang-format: Invalid argument
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -6,7 +6,6 @@ AlignConsecutiveDeclarations: None
 AlignConsecutiveMacros: None
 AlignEscapedNewlines: DontAlign
 AlignOperands: DontAlign
-AlignTrailingComments: True
 AllowAllArgumentsOnNextLine: False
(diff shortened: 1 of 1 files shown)


vabold:
Relevant comment: https://github.com/stblr/mkw-sp/issues/829#issuecomment-1581391156

GnomedDev:
Sure, but the C++ VScode extension seems to be bundling clang-format 16, and the option doesn't change how formatting on SP is done now. (see: I haven't had to touch CI and it's passing).

riidefi:
If no `BasedOnStyle` is selected, clang-format defaults to the LLVM style. The LLVM style has `AlignTrailingComments` as "true"; therefore, we can simply delete this line and avoid all of this headache.

stblr (approved):

PR #845, opened 2023-06-17
Improve dol/rel error messages

### LINKED ISSUE #755: Improve the "modified dol" error message.

GnomedDev:
This makes it simple for a user that is not familiar with Wii game patching to understand what is wrong, useful if they think the Wiimmfi patcher is a magic black box.

Closes #755

### DIFF
--- a/loader/Loader.cc
+++ b/loader/Loader.cc
@@ -123,9 +123,7 @@ std::optional<Apploader::GameEntryFunc> Run() {
         return {};
     }
     if (!*isDolClean) {
-        Console::Print(
-                "Please ensure that the file 'main.dol' is not modified\n"
-                "in any capacity!");
+        Console::Print("Please ensure the game disk is not modified in any capacity!");
         return {};
     }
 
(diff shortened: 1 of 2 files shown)


vabold:
In its current state, this isn't really an improvement.

In my view, a good error message should strive to do three things:
1) Inform the user in a way anyone can understand about what the problem is.
2) Explain what the user might have done that could have caused it.
3) Provide steps on how to potentially fix the problem.

It's true that end users likely do not know what `main.dol` and `StaticR.rel` are, but for people who aren't familiar with Wii game patching, this still doesn't provide a lot of help.

riidefi:
I think we could reword it as "Please re-download your .iso file; it is corrupt."

Zachruff:
Agree with what vabold said, also maybe change "game disc" to "game files" instead? also idk about what rii said that might be too piracy-ish with the word "redownload"

GnomedDev:
Fair, I understand it doesn't explain how to fix it, but the more text we put on screen, the less likely people are going to be to read it..

riidefi:
Have we considered color-coding our messages? We could make this message red to stand out, or something.

Zachruff:
> Have we considered color-coding our messages? We could make this message red to stand out, or something.

is that possible on the console text view? that is probably a good idea

PR #849, opened 2023-06-18
Fix gdb_compatible build

riidefi:


stblr (changes requested):
This would break the updater (see 9062cc5dc643a96aa015c7b7a984950e17285d30).

riidefi:
> This would break the updater (see https://github.com/mkw-sp/mkw-sp/commit/9062cc5dc643a96aa015c7b7a984950e17285d30).

Resolved in the latest push

stblr (approved):

PR #841, opened 2023-06-16
Implement block for Single Player VS race with no CPUs

GnomedDev:
Closes #751

PR #850, opened 2023-06-18
Implement player count setting

### LINKED ISSUE #751: VS Mode Crash
image

Using these settings:
image
image

### LINKED ISSUE #758: CPU count setting

GnomedDev:
Closes #758, closes #751  
Implements a "total player" setting, instead of a "CPU count" setting, to avoid weirdness when interacting with multiplayer.

This also re-removes the `CPUMode::None` setting, as multiplayer can simply set the player count to the number of local players, or less, allowing it to supersede #841.


Zachruff:
should this be called "total racers" instead? since i feel like "players" implies human players not CPUs

also "The minimum player count, padded by CPU players." why is it "minimum" isnt it total? and maybe im being weird but I feel like padded is a weird term. Maybe simpler to say "Total number of combined CPU and human racers"

vabold:
"Minimum" is likely a reference to the fact that it's the lower bound of local players and total players.

"Racer Count" is a good name for the setting, and I think the description can be e.g. "Guarantees at least __ racers in a race."

Zachruff:
> "Minimum" is likely a reference to the fact that it's the lower bound of local players and total players.
> 
> "Racer Count" is a good name for the setting, and I think the description can be e.g. "Guarantees at least __ racers in a race."

ah yes, this is good way of phrasing it

GnomedDev:
Yeah, it's minimum as the setting could be set to `2` but with 4 local players, it would still have all 4 local players. That name and description does sound saner though.

GnomedDev:
Actually, looking at it, I'd have to duplicate the description 12 times to name the player count, or implement new logic in the settings system which I am very adverse to doing.

vabold:
> Actually, looking at it, I'd have to duplicate the description 12 times to name the player count, or implement new logic in the settings system which I am very adverse to doing.

I think there's a way to format the string with a numerical value.

GnomedDev:
Sure, but as I said, I'd then have to mess with the settings system for one setting.

GnomedDev:
Before anyone criticizes "Gamer Count", I would like to redirect blame to Vabold who came up with the idea and repeatedly told me in DMs to commit it.

vabold:
> Before anyone criticizes "Gamer Count", I would like to redirect blame to Vabold who came up with the idea and repeatedly told me in DMs to commit it.

Strange, I don't remember any of these DMs.

We can use "Player" due to the consistency with teams.

payload/game/system/RaceConfig.cc
```diff
@@ -38,11 +38,27 @@ bool RaceConfig::isSameTeam(u32 p0, u32 p1) const {
...
-    m_menuScenario.players[0].type = Player::Type::Local;
+void RaceConfig::applyPlayers() {
+    auto *saveManager = System::SaveManager::Instance();
 
-    for (u32 i = 1; i < 12; i++) {
-        m_menuScenario.players[i].type = otherType;
+    u8 minPlayerCount = 1;
+    if (m_menuScenario.gameMode == GameMode::OfflineVS) {
+        auto setting = saveManager->getSetting<SP::ClientSettings::Setting::VSPlayerCount>();
+        minPlayerCount = static_cast<u8>(setting);
```
stblr, line 47: Is this cast needed?

assets/message/MenuSP_U.bmg.json5
```diff
@@ -1109,6 +1109,24 @@
...
+    },
+    "10434": {
+        string: "Player Count",
+    },
+    "10435": {
+        string: "Guarantees at least {arg integer|0 0} player in a battle."
+    },
+    "10436": {
+        string: "{arg integer|0 0} players",
+    },
```
stblr, line 1129: Please handle singular/plural like e.g. 10421.

stblr (changes requested):

assets/message/MenuSP_U.bmg.json5
GnomedDev: Ah, I need to make 10435 `players`, but we don't need to care about singular since the setting is minimum `2` racers/players.

payload/game/system/RaceConfig.cc
GnomedDev: Ah, not anymore.

assets/message/MenuSP_U.bmg.json5
stblr: Ah, right.

GnomedDev:
Fixed conflicts and comments.

assets/message/MenuSP_U.bmg.json5
```diff
@@ -1112,6 +1112,24 @@
...
         string: "HUD",
     },
+    "10432": {
+        string: "Racer Count",
+    },
+    "10433": {
+        string: "Guarantees at least {arg integer|0 0} racers in a race."
+    },
+    "10434": {
+        string: "{arg integer|0 0} players",
```
stblr, line 1122: Should be "racers"?

GnomedDev: Fixed

vabold (commented):
Do we want to use "racer" and "player" separately? I'd like to hear more opinions before merge.

Zachruff:
> Do we want to use "racer" and "player" separately? I'd like to hear more opinions before merge.

imo racer should be both human and CPUs, whereas player is just humans (since nintendo will refer to people as player 1, player 2, etc... and on the title screen you select "3P" mode for 3 player local for example)

riidefi:
I think it'd be more natural to use "player" even when not referring to physical humans, e.g. "Player Count" rather than "Racer Count". For instance, when we think of "24 player" mode, we aren't picturing 24 little screens as if the game were for a mosquito.

GnomedDev:
If I change to player, will this be merged?

riidefi:
I don't see any other issues--unless @MikeIsAStar has something to say.

PR #851, opened 2023-06-18
Remove total category page display from SettingsPage

### LINKED ISSUE #409: The (1/2) makes it look like there are only 2 settings pages in total
image

change it to something else (or just add vertical scrolling so you can have all the race settings on one page)

GnomedDev:
Closes #409


vabold:
I'd like to have more discussion on how to proceed with this from a UX standpoint.

GnomedDev:
I believe the medium/long term solution should be Zach's new settings page work, this is simply for fixing the short term confusion.

PR #825, opened 2023-06-04
Document more of RaceScene

riidefi:

PR #855, opened 2023-06-19
Document `EGG::ExpHeap::GroupSizeRecord`

riidefi:
Broken off #825


payload/egg/core/eggExpHeap.hh
```diff
@@ -6,6 +6,36 @@ namespace EGG {
...
+    /// The `GroupSizeRecord` class is used for tracking memory allocations for different allocation
+    /// groups. It serves as a memory budgeting system that assists in debugging out-of-memory (OOM)
+    /// crashes by recording how memory is allocated across different groups. This approach allows
+    /// precise tracking of memory usage and can help identify where the memory budget is being
+    /// exceeded.
+    ///
+    /// The class uses an array `entries[256]` where each entry corresponds to a specific allocation
+    /// group. The groups are defined in the `RKHeapTag` enum. Each time an allocation is made using
+    /// the `GameScene::setHeapTag` function, the corresponding entry in `GroupSizeRecord` is
+    /// updated to reflect the new total memory usage of that group.
```
vabold, line 18: Does this enum already exist, and if not, can we create it somewhere? Combined, this seems like it could be a useful resource for debugging crashes, given that we introduce more memory allocation.

PR #862, opened 2023-06-20
Handle key import error

stblr:


Zachruff:
all 4 region discs load on an NTSC-U Wii U

common/IOS.cc
```diff
@@ -288,9 +296,9 @@ static void Copy(const u32 (&mask)[N], const u32 (&pattern)[N], u32 *ptr) {
     }
 }
 
-bool ImportNewCommonKey() {
+std::optional<std::optional<u32>> ImportNewCommonKey() {
```
riidefi, line 299: What is the tri-state here?

riidefi:
Zachruff has also tested this on all four regions.

PR #867, opened 2023-06-21
Play a sound when a screenshot is taken

MikeIsAStar:


vabold:
To prevent a crash, the current state of the PR checks to see if `SectionManager` is initialized. Instead, I think a solution where `SectionManager` and other important REL classes are guaranteed to exist is preferable. This means creating the class somewhere in `RootScene::enter`. Given this is not the focus of the PR, I will work on infrastructure to allow this independently while Star is on vacation.

riidefi:
Given the game crashing randomly is a fairly substantial blocker for release, I am going to merge Star's workaround until a more rigorous system can be devised. The SFX code looks good to me.

PR #870, opened 2023-06-23
[StackCanary.cc] Use C++-style casts

MikeIsAStar:


payload/sp/security/StackCanary.cc
```diff
@@ -134,12 +136,16 @@ static u32 *FindLast(u32 *startAddress, u32 *endAddress, Find find, u32 instCoun
 }
 
 static u32 CreateBranchLinkInstruction(u32 *sourceAddress, u32 *destinationAddress) {
-    return (18 << 26) | (((u32)destinationAddress - (u32)sourceAddress) & 0x3FFFFFC) | (1 << 0);
+    // clang-format off
+    return (18 << 26) |
+           ((reinterpret_cast<char *>(destinationAddress) - reinterpret_cast<char *>(sourceAddress)) & 0x3FFFFFC) |
+           (1 << 0);
+    // clang-format on
```
vabold, line 143: Instead of disabling clang-format, it would be preferable if we could split out the `reinterpret_cast` calls into their own variables. This should produce a nicer looking 3-line function.

vabold (changes requested):

payload/sp/security/StackCanary.cc
MikeIsAStar: Resolved by commit [f40f4dc](https://github.com/mkw-sp/mkw-sp/pull/870/commits/f40f4dc985a5223986550f4f42e706e488bbfb56).

PR #874, opened 2023-06-25
Fix bmg-cleaner line endings

stblr:


GnomedDev:
Huh, I didn't think this was actually an issue because git was complaining but stopped when I made them CRLF. Could this be fixed in `.gitattributes`?

stblr:
What were the complaints?

GnomedDev:
`warning: in the working copy of 'assets/message/CommonSP_D.bmg.json5', LF will be replaced by CRLF the next time Git touches it`

This does only occur when on Windows using VSCode/git inside WSL, so it could possibly be a me problem.

PR #877, opened 2023-06-26
Add infra for loading custom tracks over an existing slot

GnomedDev:
This allows important course specific parameters to be replaced to no longer depend on the `courseId` field, allowing for the core engine code (`RaceConfig`) to stay separate from the Track Packs implementation.


GnomedDev:
Ready for review.

payload/sp/storage/DecompLoader.cc
```diff
@@ -29,13 +30,22 @@ static OSThread thread;
 alignas(0x20) static u8 srcs[2][0x20000 /* 128 KiB */];
 
 static std::optional<FileHandle> Open(const char *path, std::optional<StorageType> storageType) {
-    if (ThumbnailManager::IsActive()) {
+    auto *raceConfig = System::RaceConfig::Instance();
+
+    // This is called before the game is loaded, so the nullptr check is actually needed.
+    if (raceConfig != nullptr && raceConfig->m_spRace.pathReplacement.m_len != 0) {
```
riidefi, line 36: This is super messy. Is there anything preventing us pulling the new code out, so DecompLoader doesn't have a circular dependency with the game logic? I'd further prefer if the ThumbnailManager code was also yanked, but I am fine with that being saved for another PR.

riidefi (changes requested):

riidefi:
Thanks, and nice work. Only one thing I spotted with the path replacement and then we should be good to merge!

payload/sp/storage/DecompLoader.cc
GnomedDev: The problem is that the nice patch I did to ResourceManager::loadCourse doesn't work anymore because `ro:/` was patched to not read from the SD, and the track path that loadCourse sets up gets prepended with `ro:/`. 

Other places to patch would require rewriting more of ResourceManager's internals which I'm not too familiar with, or reverting the `ro:/` commit (I don't know which one it is) which would be confusing.

I agree this is ugly, but it is the best place we have right now and has no functional downside.

PR #881, opened 2023-06-29
Implement basic Track Pack System

GnomedDev:
A re-implementation of the "unmergable" #655.

This track pack system builds on the replacement infrastructure built in #877 by reading from "Track Pack" manifests found in `/mkw-sp/Track Packs` which contain basic metadata and lists of Sha1 hashes for track files. Track metadata is then loaded from `/mkw-sp/Tracks`, which also contains the actual track files as `.arc.lzma`.

This is all managed by `TrackPackManager`, which is loaded on sections which require pack selection, then persists by using the `courseOrder` arrays introduced in #805 to store `Track` structs, which when needed are applied to `RaceConfig`.

Further work is planned to implement a server which would serve track/pack metadata files and tracks, as well as UI to interact with this server, but currently this is purely offline and requires messing around with [Track Pack Generator](https://github.com/GnomedDev/track-pack-generator) to be usable, although this can currently convert CTGP and other packs to this format. `Wiimm-DB-Parser` is included in this PR to produce files for this tool, and in future form the basis for the server, but this could be removed and maintained separately if wanted.


GnomedDev:
Ready for review, should be reviewable commit by commit.

PR #807, opened 2023-06-01
Add dummy PackSelectPage

GnomedDev:
Split from #655. Currently does nothing, gets in the way of getting to a race, and crashes online.

PR #883, opened 2023-06-30
Fix PackSelectPage online asset issues

GnomedDev:
This allow the online UI to load again after #807. It also fixes some other small online UI issues related to track packs.

### DIFF
--- a/payload/game/ui/OnlineTopPage.cc
+++ b/payload/game/ui/OnlineTopPage.cc
@@ -102,6 +102,17 @@ void OnlineTopPage::onWorldwideButtonFront(PushButton *button, u32 /* localPlaye
 }
 
 void OnlineTopPage::onTrackpackButtonFront(PushButton *button, u32 /* localPlayerId */) {
+    // TODO(GnomedDev): Track Pack support
+    auto packCount = 1;
+    if (packCount == 1) {
+        auto *section = SectionManager::Instance()->currentSection();
+        auto messagePage = section->page<PageId::MessagePopup>();
+
+        messagePage->reset();
+        messagePage->setWindowMessage(20047);
+        return push(PageId::MessagePopup, Anim::None);
+    }
+
     m_replacement = PageId::PackSelect;
     startReplace(Anim::Next, button->getDelay());
 }
--- a/payload/game/ui/PackSelectPage.cc
+++ b/payload/game/ui/PackSelectPage.cc
@@ -8,6 +8,7 @@
 #include "PackSelectPage.hh"
 
 #include "game/ui/CourseSelectPage.hh"
+#include "game/ui/MessagePage.hh"
 #include "game/ui/ModelPage.hh"
 #include "game/ui/OnlineConnectionManagerPage.hh"
 #include "game/ui/SectionManager.hh"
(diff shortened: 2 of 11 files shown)


payload/game/ui/OnlineTopPage.cc
```diff
@@ -102,6 +102,16 @@ void OnlineTopPage::onWorldwideButtonFront(PushButton *button, u32 /* localPlaye
 }
 
 void OnlineTopPage::onTrackpackButtonFront(PushButton *button, u32 /* localPlayerId */) {
+    auto packCount = 1;
```
vabold, line 105: I would recommend commenting that this a temporary initialization.

payload/game/ui/PackSelectPage.cc
```diff
@@ -185,9 +189,9 @@ void PackSelectPage::refresh() {
     auto sectionId = SectionManager::Instance()->currentSection()->id();
     bool isOnline = Section::HasOnlineManager(sectionId);
 
-    u32 packCount = 1 - isOnline;
-    for (size_t i = isOnline; i < m_buttons.size(); i++) {
-        u32 packIndex = m_sheetIndex * m_buttons.size() + i;
+    u32 packCount = 1;
```
vabold, line 192: Same here.

vabold (commented):
Note that these are just comments that I think will help communication in development. You're free to close them if you want.

payload/game/ui/PackSelectPage.cc
```diff
@@ -110,11 +114,11 @@ void PackSelectPage::onButtonFront(PushButton *button, u32 /* localPlayerId */)
     auto *section = SectionManager::Instance()->currentSection();
     auto *connManager = section->page<PageId::OnlineConnectionManager>();
 
+    auto buttonIndex = m_sheetIndex * m_buttons.size() + button->m_index;
     if (connManager != nullptr) {
-        connManager->setTrackpack(button->m_index * m_sheetIndex);
+        connManager->setTrackpack(buttonIndex + 1);
```
vabold, line 119: Based on existing implementation, it seems as though `setTrackpack` is meant to set the global pack ID, but the call here is only local. For example, what happens if we have track packs 2 and 4, but the button index suggests we should be setting track pack 3 (even though we don't have track pack 3)?

vabold (changes requested):

payload/game/ui/PackSelectPage.cc
GnomedDev: There are no stable IDs, so this can't work in practice. I implemented this "track pack" stuff back with the initial matchmaking PR as scaffolding to build around. When we have a pack server and a stable ID system this could be properly fixed.

vabold: Okay, sounds good. I think we could add a comment to reflect that (we love these here).

GnomedDev: I think due to the open nature of #881 and this discussion to link back to if it ever gets mentioned again (which I doubt, seeing it's both my things, online UI and track packs), it wouldn't be worth it to fuss over a comment which would require discussion on where to place.

payload/game/ui/PackSelectPage.cc
GnomedDev: This is already documented in the top level file comment.

payload/game/ui/OnlineTopPage.cc
GnomedDev: Done

PR #886, opened 2023-07-04
Validate '.tpl' files

MikeIsAStar:


vabold (approved):
The purpose of this PR is to prevent pointer writes to file-defined offsets in unwanted parts of memory. It's unclear if we have access to secure file sizes for any given TPL, so I think validating the header like this is fair.

This is a good local security measure and the asserts haven't popped up yet in brief testing. It could theoretically be possible for this offset to not be 0xC and still be a valid TPL, therefore causing false positives. However, this is only a theoretical possibility, and I think it's unlikely this will be the case in the future.

I approve, but I think having another set of eyes on this before merging is a good idea.

payload/revolution/tpl.cc
```diff
@@ -0,0 +1,11 @@
+extern "C" {
+#include "tpl.h"
+}
+
+extern "C" void TPLBind(TPLPalette *tpl) {
+    constexpr u32 IMAGE_DESCRIPTOR_ARRAY_OFFSET = 0x0000000C;
+
+    assert(tpl->numImageDescriptors > 0);
```
riidefi, line 8: It should probably remain possible to bind an empty `.tpl` archive.

riidefi (changes requested):
Re: Array offset -- this is fine. Doing it properly would require changing the API of TPLBind which is not worth the effort.

payload/revolution/tpl.cc
MikeIsAStar: Resolved by commit [177272f](https://github.com/mkw-sp/mkw-sp/pull/886/commits/177272f28b31d72a9053a482e63ce535ca0bd937).

PR #894, opened 2023-07-16
Add setting for the binding of Y button

### LINKED ISSUE #893: Setting for binding of Y button
Currently, the Y button is allocated to two candidate features, Screenshots and Item Menu. This is problematic and blocks merging of the latter. A simple setting selecting between the two will resolve this before more extensive controller support comes.

riidefi:
Closes #893. Adds a setting to multiplex features potentially sharing this button.


GnomedDev (approved):
Hmm, I'm uneasy with explicitly calling it the "Y" button, in case we support more controllers such as wiimote, but as is this is good.

PR #905, opened 2023-07-21
Hash battle tracks alongside the race tracks

GnomedDev:


riidefi:
How long does this take on console?

GnomedDev:
I haven't checked, but possibly a significant amount of time. Having just thought about it, I could remove the disk hash patch if I implement more logic down the road for #884.

GnomedDev:
Just checked and it takes an extra 30 seconds on boot, which is unacceptable, so I have removed that section of the PR.

PR #908, opened 2023-07-21
Add support for WBZ track files

### LINKED ISSUE #884: Support reading .wbz files directly

GnomedDev:
Had to be duplicated due to a Github bug. See #890. Closes #884.


payload/sp/WU8Library.hh
```diff
@@ -0,0 +1,29 @@
...
+    Ripping,
+    Decompressing,
+    Processing,
+    Writing,
+    Finished,
+    ReplacedCourse,
+};
+
+void StartExtraction(EGG::Heap *mem2);
+std::tuple<ExtractionState, std::array<char, 64>> GetExtractionState();
```
riidefi, line 22: I'm really not a fan of `std::pair<>` and `std::tuple` over just a struct with two named fields.

payload/sp/WU8Library.cc
```diff
@@ -0,0 +1,536 @@
...
+    constexpr size_t CompressedMaxSize = 3000000;
+    constexpr size_t DecompressedMaxSize = 6000000;
+
+    void *compressedBlock = extractionHeap->alloc(CompressedMaxSize, -0x20);
+    void *decompressedBlock = extractionHeap->alloc(DecompressedMaxSize, -0x20);
+    std::span<u8> compressedBuf(reinterpret_cast<u8 *>(compressedBlock), CompressedMaxSize);
+    std::span<u8> decompressedBuf(reinterpret_cast<u8 *>(decompressedBlock), DecompressedMaxSize);
+
+    u32 u8MagicInt = 0;
+    memcpy(&u8MagicInt, U8_MAGIC, 4);
```
riidefi, line 344: Do we need this?

payload/sp/WU8Library.cc
```diff
@@ -0,0 +1,536 @@
...
+    OSResumeThread(&s_extractionThread);
+    OSDetachThread(&s_extractionThread);
+}
+
+void *ExtractThread(void *param) {
+    auto *extractionHeap = reinterpret_cast<EGG::ExpHeap *>(param);
+    assert(Storage::CreateDir(L"autoadd/", true));
+
+    constexpr size_t CompressedMaxSize = 3000000;
+    constexpr size_t DecompressedMaxSize = 6000000;
```
riidefi, line 336: Add some comment explaining these values. Should also be noted that CTGP supports up to 8MB decompressed, if we want parity (e.g. Autumn Leavesway).

payload/sp/WU8Library.hh
GnomedDev: In my eyes, it is clear enough what is going on, and a struct would just be a waste of time.

payload/sp/WU8Library.cc
```diff
@@ -0,0 +1,536 @@
...
+
+    ExtractionState state = s_extractionState;
+    auto currentlyExtracting = s_currentlyExtracting;
+
+    OSUnlockMutex(&s_extractionStateLock);
+    return std::make_tuple(state, currentlyExtracting);
+}
+
+void SetExtractionState(ExtractionState state, const char *currentFile) {
+    OSLockMutex(&s_extractionStateLock);
```
riidefi, line 301: I think we can use `std::unique_lock`, `std::scoped_lock` or `SP::ScopeLock` here.

payload/sp/WU8Library.cc
GnomedDev: Eh, prevents having another magic number.

riidefi: I mean the `memcpy`. Both are of type "u32", so a simple assignment should work (?)

payload/sp/WU8Library.cc
GnomedDev: `ScopeLock` looks like it locks when constructed? I'm not sure, I tried not over engineering this and keeping it as simple as possible to prevent threading bugs.

payload/sp/WU8Library.cc
GnomedDev: Nope, `U8_MAGIC` is a 4 byte array, you are getting mixed up with `WU8_MAGIC`.

payload/sp/WU8Library.cc
GnomedDev: Done

payload/sp/WU8Library.cc
GnomedDev: Done

GnomedDev:
Only problem I currently see is that while the speed is uncapped, music still plays, but this can be smoothed out in future when some infra has been implemented that isn't so jank.

payload/sp/WBZDecoder.cc
```diff
@@ -0,0 +1,115 @@
...
+
+bool WBZDecoder::done() const {
+    return m_done;
+}
+
+bool WBZDecoder::CheckMagic(u64 magic) {
+    return magic == WBZ_MAGIC;
+}
+
+bool WBZDecoder::isOutputFull(int retCode, std::span<char> decodeBuf) {
```
riidefi, line 95: This seems like a somewhat misleading name?

payload/sp/WBZDecoder.cc
```diff
@@ -0,0 +1,115 @@
...
+#include <cstring>
+
+// WBZaWU8a
+constexpr u64 WBZ_MAGIC = 6287687403886164065;
+
+namespace SP {
+
+void *WBZDecoder::Alloc(void *heap_erased, int item, int count) {
+    auto *heap = reinterpret_cast<EGG::Heap *>(heap_erased);
+    return heap->alloc(item * count, 0x4);
```
riidefi, line 21: Perhaps this should be -0x4 to avoid allocation interleaving / contention between BZ2 and the std::vector?

payload/sp/WU8Library.cc
```diff
@@ -0,0 +1,582 @@
...
+
+    s_extractionStage = stage;
+    if (currentFile == nullptr) {
+        s_currentlyExtracting[0] = '\0';
+    } else {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
+        strncpy(s_currentlyExtracting.data(), currentFile, s_currentlyExtracting.size());
+        s_currentlyExtracting[s_currentlyExtracting.size() - 1] = '\0';
+#pragma GCC diagnostic pop
```
riidefi, line 311: Could this be cleaned up by using snprintf? The perf here shouldn't matter (and for that matter GCC should be able to transform snprintf to this code anyway)

payload/sp/WU8Library.cc
```diff
@@ -0,0 +1,582 @@
...
+        }
+
+        auto nameLen = wcslen(node->name);
+        if (nameLen <= 4) {
+            continue;
+        }
+
+        if (!wcscmp(node->name + nameLen - 4, L".wbz")) {
+            return true;
+        }
```
riidefi, line 341: `std::wstring_view(node->name).ends_with(".wbz")`

payload/game/ui/ctrl/CtrlBusySymbol.hh
```diff
@@ -0,0 +1,13 @@
...
+
+#include "game/ui/UIControl.hh"
+
+namespace UI {
+
+class BusySymbol : public LayoutUIControl {
+public:
+    BusySymbol();
+    virtual ~BusySymbol();
+};
```
riidefi, line 11: `static_assert` that `sizeof(BusySymbol)` is as-expected.

payload/sp/WBZDecoder.cc
GnomedDev: I actually have no idea what the `-` is doing in align lmao.

payload/sp/WU8Library.cc
```diff
@@ -0,0 +1,582 @@
...
+    auto myStuff = SP::Storage::OpenDir(L"My Stuff");
+    if (!myStuff) {
+        return false;
+    }
+
+    return ContainsWBZ(*myStuff);
+}
+
+void StartExtraction(EGG::Heap *mem2) {
+    size_t size = OSRoundUp32B(10000000);
```
riidefi, line 361: Maybe
```cpp
// 1MB for BZ2
constexpr size_t ExtractionHeapSize = CompressedMaxSize + DecompressedMaxSize + 1'000'000;
```

payload/sp/WBZDecoder.cc
riidefi: If you think of the heap as a giant tape, positive alignment objects are allocated from the left, and negative alignment value objects are allocated from the right. They meet in the middle somewhere.

payload/sp/WBZDecoder.cc
GnomedDev: This is black magic, I played about until it worked. If this returns true, it means that the decode loop should continue, because bz2 has run out of output space, so that's what I named it.

riidefi: `processDecodedPacket` maybe

GnomedDev:
Solved all review comments.

riidefi (approved):
LGTM, thanks!

PR #924, opened 2023-08-11
Try mount SD first(?)

Naim2000:
This one feels kind of funny because. It took me 18 tries for mkw-sp to pick USB instead of SD 5 times. Of course with this single line edit, it would mount my USB a grand total of 0 times out of those same 18. However i'm just surprised that it mounted my USB only... 5 times. I guess it's just Dolphin


Naim2000:
This also might end up fixing https://github.com/mkw-sp/mkw-sp/issues/572

PR #925, opened 2023-08-13
Oversight patches in TimeAttackGhostListPage, use intended sound functions instead of playSound

JohnP55:


JohnP55:
Fixes:
- Selecting a track always makes the ghost select sound
- Hovering over the "watch" button and going to the course select and back makes the settings button unreachable
- Selecting a ghost and going to the course select and back hovers over a non-existent button (the Race button while no ghosts are active)

PR #931, opened 2023-08-23
Allocate the heap used by the 'NHTTP' library in MEM2

MikeIsAStar:


riidefi:
This looks good. My only concern would be if the NHTTP buffer size might change depending on the nature of the request we make (even how the server responds), but if it covers all cases this seems fine.

I don't have the rankings server set up to verify--I trust you have done that here? Ideally it would be best if you could PR that code sooner than later to make testing these a bit easier.

